PR #4, opened 2022-11-17
Bump dependencies in opam file

jesyspa:
Hi, a clean checkout of the repository currently doesn't build (stdpp has moved `numbers.Qp` into a module, amongst other things), so I've updated the dependencies to the latest versions on opam.  This builds and `make test` passes.

(The update to `coq-ext-lib` is not necessary, it builds and passes with the old version as well, but I thought it may be worth doing for consistency.)

### DIFF
--- a/coq-cpp2v.opam
+++ b/coq-cpp2v.opam
@@ -12,11 +12,11 @@ bug-reports: "https://github.com/bedrocksystems/cpp2v/issues"
 dev-repo: "git+https://github.com/bedrocksystems/cpp2v.git"
 
 depends: [
-  "coq" { = "8.15.1" }
-  "coq-ext-lib" { = "0.11.6" }
-  "coq-iris" {(= "dev.2022-04-12.0.a3bed7ea")}
-  "elpi" {(= "1.15.2")}
-  "coq-elpi" {(= "1.14.0")}
+  "coq" { = "8.16.0" }
+  "coq-ext-lib" { = "0.11.7" }
+  "coq-iris" {(= "dev.2022-11-16.2.da02a192")}
+  "elpi" {(= "1.16.7")}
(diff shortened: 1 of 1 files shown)


gmalecha:
Thanks, @jesyspa! We're looking into it. The current opam file should work, but it is probably different than what you had installed on your system.

Blaisorblade:
Thanks a lot for the report! I think @jesyspa is right and the current opam file does _not_ work, because it was not used by our internal CI testing.

jesyspa:
If it helps, one of the errors was [here](https://github.com/bedrocksystems/BRiCk/blob/master/theories/prelude/numbers.v#L495-L497): `Qp` only became a module in [this commit of stdpp](https://gitlab.mpi-sws.org/iris/stdpp/-/commit/56af224a44aecaea36a95415f96b95e52002e56d).
(I used `opam install coq-cpp2v coq-cpp2v-bin` to build, and I think that that downgraded the packages accordingly.)

gmalecha:
Thanks. I misunderstood the issue.

Blaisorblade:
Thanks — I think this should be fixed now, even if we could not take your patch.

- https://github.com/bedrocksystems/BRiCk/commit/3fe5fefc777e97865520e9a0285c15352adbc2a0 extends CI to test the opam packages going forward — I hope I credited you appropriately @jesyspa 😄 .

- https://github.com/bedrocksystems/BRiCk/commit/906c51a4bb2a5fd2eed28fb438fc4d5509d56807 also contains the opam updates. I've used the versions we were already testing internally — there's nothing wrong with your choices, but a new bump would affect additional infrastructure.

jesyspa:
Awesome, thanks, I've checked and it works out of the box now! :)

PR #7, opened 2024-12-14
Fix typos and grammar: README.md

CharlesCNorton:
-Corrected "only recompiling files that need to be," to "only recompile files that need to be," for grammatical accuracy.

-Fixed "no lot is included" to "no log is included" to resolve a typo.


gmalecha:
Thanks!

PR #8, opened 2025-01-07
Start Typed Pointers

ehatti:
The goal is to create a simpler model than `inductive_pointers.v` that still admits the pointer axioms.

If my knowledge is correct, two pointers are equal when 1. they have the same allocation ID, 2. they have the offset within that allocation, and 3. they point to subobjects of the same type. So this model tracks almost those: the offset within the base type of the allocation, the type at the base of that offset, and the type at the end of that offset. I have yet to add in the allocation ID.

It then also contains a side condition that requires the offset to trace a path in the base type to the destination type. This goal here is that this side condition contains all the type information and such needed to prove `VALID_PTR_AXIOMS`. And since it is a `Prop`, the operations on pointers are simple -- they're just offset calculations + proofs of the side condition, which has easy equality since it's in `Prop`.

One issue: This model requires `offset` to be indexed by a `genv`. Is it reasonable to add this to the theory, or should the model be reworked?


rocq-bluerock-brick/theories/lang/cpp/model/typed_pointers.v
```diff
@@ -0,0 +1,165 @@
...
+#[local] Close Scope nat_scope.
+#[local] Open Scope Z_scope.
+
+Import canonical_tu.
+Module LAYERED_PTRS_IMPL <: PTRS_INTF.
+  Variant wf_offset_seg `{σ : genv} : type -> Z -> type -> Prop :=
+  | WfSub ty i n s :
+      Some s = size_of σ ty ->
+      (i < n)%N ->
+      wf_offset_seg (Tarray ty n) (i * s) ty.
```
Blaisorblade, line 24: Not sure what are the types of `i` and `n`, but `o_sub σ ty (-2)` is a valid pointer offset, and one that isn't applicable to complete arrays.

gmalecha:
Thanks for the hacking @ehatti! Some notes from our meeting:

```
struct C {
    int x[1];
    int y[1];
};

void test() {
  C c;
  (c.x + 1) == (c.y)
  
  c.x[1] = 0; // UB
  c.y[0] = 0; // legal
  
  *((c.x + 1) - 1) = 0; // legal
  *((c.x + 2) - 2) = 0; // UB
}
  
p is a valid pointer ->
p ., o_derived A B ., o_base B A = p ->
p ., o_derived A B != structurally invalid



raw          (type)
canonical    (type)
c++ validity (_ -> state -> Prop)
```

PR #9, opened 2025-01-20
String rewrite system based definition of offset normalization

ehatti:
Redefines the definition of pointer offset normalization in [this file](https://github.com/ehatti/BRiCk/blob/srw-ind-pts/rocq-bluerock-brick/theories/lang/cpp/model/inductive_pointers2.v) as a string rewrite system. Current progress: The pointer equalities in the `PTRS` interface are all provable, `vaddr` and `alloc_id` axioms have not been proven -- those are orthogonal to the definition of normalization however.

`roff_rw_local` defines the local rewrites, namely that `o_base` and `o_derived` cancel out, that indexing the 0th element is a no-op, and that sequences `o_sub`s should be merged together. `roff_rw_global` lifts this to a rewrite relation on entire offsets, and finally `roff_rw` takes the closure of this to define the rewrite system.

From there we define a function `normalize` -- which is used to ensure the result of offset composition is still normalized -- and prove it is equivalent to `roff_rw`. I've proven that it is sound wrt `roff_rw`, but I still need to prove the opposite direction.

I'm using the Equations library to get nicer well-founded recursion, since `Program Fixpoint` generates quite unworkable terms. Let me know if adding this extra dependency is a problem, if it is then we can switch to a different definition for `normalize`. 
Equations is just nice because it "just works" and so makes it faster to get the idea into Coq.

PR #13, opened 2025-04-04
Add new pointer model, based on a string rewrite system

Blaisorblade:
Redefines pointer-offset normalization as a string rewrite system. We've already tweaked the `ptrs.v` interface to fit this model.

This is just a squashed version of #9 — I'm merging this internally.


rocq-bluerock-brick/theories/lang/cpp/model/inductive_pointers2.v
```diff
@@ -0,0 +1,1319 @@
+(*
+ * Copyright (c) 2020 BlueRock Security, Inc.
```
Blaisorblade:
```suggestion
 * Copyright (c) 2020-2025 BlueRock Security, Inc.
```

PR #14, opened 2025-05-05
port to elpi 3.0

gares:
This is a port to the (upcoming) elpi 3.0. The main difference is that the language feature a static check for determinacy, i.e. code that uses no backtracking. The main change here is that derive wants derivations to be functions (not relations), hence I had to declare them as such.

It uncovered a little bug in Elpi, hopefully I'll manage to integrate this repo in Elpi's CI
https://github.com/LPCIC/elpi/pull/337

### DIFF
--- a/rocq-bluerock-prelude/theories/elpi/derive/bitset.v
+++ b/rocq-bluerock-prelude/theories/elpi/derive/bitset.v
@@ -25,7 +25,6 @@ Elpi Db derive.bitset.db lp:{{
   pred finite-type-done o:gref.
   pred bitset-done o:gref.
 }}.
-Elpi Accumulate derive File derive.finite_type.elpi.
 Elpi Accumulate derive lp:{{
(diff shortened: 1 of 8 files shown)


Janno (commented):
Thanks for porting our code. I am only now getting around to bumping rocq and coq-elpi. The thing that has me a bit worried here is the removal of `std.do!` which we use very frequently. Is there a reason it shouldn't be used `func`s?

gares:
No, you can keep it.
Especially to be able to turn it into `std.spy-do!`.
But semantically it is unnecessary since the static checker roughly ensure that each call to a functional `p` is equivalent to `once p` where `once P :- P, !.`, that is what `do!` does for you (at runtime), so the `!` added by `do!` becomes useless and `std.do!` can be removed altogether.

gares:
This pr is now outdated, I think we can close it.

PR #28, opened 2025-11-04
[docs] separately apply header/footer.

rlepigre-skylabs-ai:
This allows enabling the dune cache for building docs, while we wait for https://github.com/ocaml/dune/pull/11131 or something similar.


Makefile
```diff
@@ -32,9 +32,13 @@ doc:
...
 	@cp -r coqdocjs/extra/resources/*.css doc/sphinx/_static/css/coqdocjs
 	@cp -r coqdocjs/extra/resources/*.js doc/sphinx/_static/js/coqdocjs
-	@ROCQLIB=${ROCQLIB} dune build --cache=disabled @doc
+	@ROCQLIB=${ROCQLIB} dune build @doc
 	@mkdir -p ${COQDOC_DIR}
 	@cp -r -t ${COQDOC_DIR} $$(find ${BUILD_ROOT} -type d -name '*.html')
+	@find ${COQDOC_DIR} -type f -name '*.html' | xargs -n 1 -P 16 -I {} sed -i \
+		-e '/{{{FOOTER}}}/{' -e 'r coqdocjs/extra/footer.html' -e 'd' -e '}' {}
+	@find ${COQDOC_DIR} -type f -name '*.html' | xargs -n 1 -P 16 -I {} sed -i \
+		-e '/{{{HEADER}}}/{' -e 'r coqdocjs/extra/header.html' -e 'd' -e '}' {}
```
pgiarrusso-sl, line 46: - Won't work on Mac, needs `gsed` there
- `sed -i` works on multiple files just fine, so no need for `-n 1 -P 16`?
- can this be written as a single `-e` argument? EDIT: even a single sed invocation

pgiarrusso-sl (approved):
Approving up to this comment.

Makefile
rlepigre-skylabs-ai: The `-n 1 -P 16` ensures that we're gonna run with up to 16 processes, as per the documentation of `xargs`, which encourages the use of `-n` with the `-P` option. But actually, `-I` implies `-n 1` I think, so this is useless (there is even a warning about it).

About the `-e` option, I tried, but could not make it work.

Makefile
```diff
@@ -22,6 +22,8 @@ BUILD_ROOT=../../_build/default/fmdeps/BRiCk
 COQDOC_DIR=doc/sphinx/_static/coqdoc
 ROCQLIB=${PWD}/../../_build/install/default/lib/coq
 
+SED = $(shell which gsed 2> /dev/null || which sed 2> /dev/null)
```
pgiarrusso-sl, line 25:
```suggestion
SED = $(shell (which gsed || which sed) 2> /dev/null)
```

rlepigre-skylabs-ai: Let's do that another time, I don't want to re-run CI now.

PR #29, opened 2025-11-05
Test case for comments in source code

gmalecha-at-skylabs:
Opened to attempt to reproduce https://github.com/SkylabsAI/auto/issues/20.

I was unable to reproduce this locally. This adds a test for CI, but it seems like it is a proof general problem.


pgiarrusso-sl (changes requested):
IMHO we should merge the MR, but it doesn't close that bug, so the description should be changed — the "closes" line should be removed, both for clarity and because GitHub applies it.

simon-skylabs:
I think Paolo is right, we can merge the test and keep going trying to chase down the bug. I'll keep interacting with the issue.

simon-skylabs (approved):

pgiarrusso-sl (approved):

Issue #25, opened 2025-11-03
Update CI to publish documentation updates


gmalecha-at-skylabs:
The BRiCk, sphinx-based, documentation should be re-published in every merge to master.

pgiarrusso-sl:
(We'll need to figure what this means, but I guess the answer is in the old CI code).

pgiarrusso-sl:
Maybe @gmalecha-at-skylabs needs also the Sphinx docs here.

rlepigre-skylabs-ai:
I started work on this in https://github.com/SkylabsAI/workspace/pull/41. Since this needs building dependencies, building the docs needs to happen as part of the workspace infrastructure (to be properly triggered).

pgiarrusso-sl:
Status: docs are now built but not yet published.
- https://github.com/SkylabsAI/BRiCk/pull/38 fixes doc builds after regressions
- https://github.com/SkylabsAI/workspace/pull/41 builds them to prevent regressions.
- publishing is still pending.

rlepigre-skylabs-ai:
Fixed by https://github.com/SkylabsAI/workspace/pull/76.

PR #38, opened 2025-11-13
Fix make doc — on Macs & elsewhere

pgiarrusso-sl:
I just wanted to fix Mac-specific breakage, but I found and fixed more problems — some being regressions from #24.

Helps #25.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,9 @@ _CoqProject: _CoqProject.template
 SHELL := /bin/bash
 
 BUILD_ROOT=../../_build/default/fmdeps/BRiCk
-COQDOC_DIR=doc/sphinx/_static/coqdoc
 ROCQLIB=${PWD}/../../_build/install/default/lib/coq
+DOC_PATH = rocq-bluerock-brick/doc
+COQDOC_DIR = $(DOC_PATH)/sphinx/_static/coqdoc
 
 SED = $(shell which gsed 2> /dev/null || which sed 2> /dev/null)
 CP = $(shell (which gcp || which cp) 2> /dev/null)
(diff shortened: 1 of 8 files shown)


rocq-bluerock-brick/README.md
```diff
@@ -29,6 +13,10 @@ using
 $ dune build
 ```
 
+### Building docs
+
+Install `python_requirements.txt` in your Python virtual environment, then run `make doc`.
```
rlepigre-skylabs-ai: How hard would it be to use `uv` as in other places, so that `make doc` takes care of the setup?

pgiarrusso-sl: I wasn't sure what to pick, but `uv run --with-requirements` should do it.

Makefile
```diff
@@ -19,40 +19,36 @@ _CoqProject: _CoqProject.template
...
-SED = $(shell which gsed 2> /dev/null || which sed 2> /dev/null)
+SED = $(shell (which gsed || which sed) 2> /dev/null)
+CP = $(shell (which gcp || which cp) 2> /dev/null)
 
 doc:
-	@dune clean
-	@dune build @../vendored/rocq/install
-	@dune build
-	@rm -rf /tmp/coqdocjs
-	@cp -r coqdocjs /tmp
```
rlepigre-skylabs-ai, line 32: How are things supposed to work without this? The `coqdoc` flags expect files in `/tmp/coqdocjs`, right?

pgiarrusso-sl: Something seems wrong indeed — I hadn't seen it. This seemed dead, and nothing broke when I removed it — maybe because of `/tmp/coqdocjs` remained.

rlepigre-skylabs-ai: Yeah. I think we could replace that code with an `echo "{{{HEADER}}}" > /tmp/...`, and remove the dummy files. This would look more deliberate.

pgiarrusso-sl: Dropped this, and pulled refactorings/cleanups out.

rocq-bluerock-brick/README.md
pgiarrusso-sl: Done.

rlepigre-skylabs-ai (approved):

Issue #40, opened 2025-11-14
creating `_CoqProject` files from `dune` files ignores symbolic links


simon-skylabs:
In workspace, `make stage1` creates a `_CoqProject` that ignores `NOVA` completely because its `dune` files are symbolic links

PR #41, opened 2025-11-14
Fix: consider dune files that are symbolic links when generating `_CoqProject` files

simon-skylabs:
https://github.com/SkylabsAI/BRiCk/issues/40


rlepigre-skylabs-ai (approved):
Looks good to me.

PR #42, opened 2025-11-14
_CoqProject generation: style revisions

pgiarrusso-sl:
Simplifications over #41.

### DIFF
--- a/scripts/coq_project_gen/gen-_CoqProject-dune.sh
+++ b/scripts/coq_project_gen/gen-_CoqProject-dune.sh
@@ -37,11 +37,11 @@ echo
 echo "# Specified logical paths for directories (for .v and .vo files)."
 
 list_dune_files () {
-    local find_args=()
-    find_args+=( \( -name _build -o -name .git \) )
-    find_args+=( -prune -false -o -name dune )
-    find_args+=( \( -type f -o -type l \) )
-    find . "${find_args[@]}" -print | sort
+  find . \
+    \( -name _build -o -name .git \) \
+     -prune -false -o -name dune \
+     \( -type f -o -type l \) \
+    -print | sort
 }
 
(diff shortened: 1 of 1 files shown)


scripts/coq_project_gen/gen-_CoqProject-dune.sh
```diff
@@ -37,11 +37,10 @@ echo
 echo "# Specified logical paths for directories (for .v and .vo files)."
 
 list_dune_files () {
-    local find_args=()
-    find_args+=( \( -name _build -o -name .git \) )
-    find_args+=( -prune -false -o -name dune )
-    find_args+=( \( -type f -o -type l \) )
-    find . "${find_args[@]}" -print | sort
+  find . \
+    \( -name _build -o -name .git \) -prune -false -o \
```
simon-skylabs: on this line, I find it unclear what the disjunction `-o` applies to. I'm assuming it's `-prune -false` on one hand and `-name dune` on the other hand. Can we put them on the same line and put brackets around them?

rlepigre-skylabs-ai (approved):

scripts/coq_project_gen/gen-_CoqProject-dune.sh
pgiarrusso-sl: I'm afraid your assumption is not correct. But, now I understand why you broke lines the way you did. I tweaked my line breaks a little more.

The arguments to `-o` are `\( -name _build -o -name .git \) -prune -false` and `-name dune \( -type f -o -type l \) -print` — as usual, disjunction has priority lower than conjunction, and conjunction is implicit.

This command-line means "if an entry is named `_build` and `.git`, don't traverse it (`-prune`) and don't even print it (`-false`); if an entry is named `dune` and is a file or symlink, print it.

PR #43, opened 2025-11-14
Cleanup `make doc`

pgiarrusso-sl:
New version of #39.


Makefile
```diff
@@ -14,50 +17,49 @@ all:
...
-	@$(CP) -r -t ${COQDOC_DIR} $$(find ${BUILD_ROOT} -type d -name '*.html')
-	@find ${COQDOC_DIR} -type f -name '*.html' \
-		| xargs -P 16 -I {} $(SED) -i \
-		-e '/{{{FOOTER}}}/{' -e 'r coqdocjs/extra/footer.html' -e 'd' -e '}' {}
-	@find ${COQDOC_DIR} -type f -name '*.html' \
-		| xargs -P 16 -I {} $(SED) -i \
-		-e '/{{{HEADER}}}/{' -e 'r coqdocjs/extra/header.html' -e 'd' -e '}' {}
-	@uv run --with-requirements python_requirements.txt $(MAKE) -C $(DOC_PATH) html
+	$(Q)rm -rf /tmp/coqdocjs
+	$(Q)cp -r coqdocjs /tmp
```
pgiarrusso-sl: I missed a suggestion here from https://github.com/SkylabsAI/BRiCk/pull/38#discussion_r2523691366

Makefile
```diff
@@ -14,50 +17,49 @@ all:
...
 
 _CoqProject: _CoqProject.template
-	@cp $< $@
+	$(Q)cp $< $@
 
 SHELL := /bin/bash
 
-BUILD_ROOT=../../_build/default/fmdeps/BRiCk
-ROCQLIB=${PWD}/../../_build/install/default/lib/coq
+dune_build_folder = ../../_build
```
rlepigre-skylabs-ai, line 24: Maybe make the variable all caps for consistency?

Makefile
rlepigre-skylabs-ai: Yeah, I think that one would be strictly better.

rlepigre-skylabs-ai (approved):

Makefile
pgiarrusso-sl: https://www.gnu.org/software/make/manual/html_node/Using-Variables.html

> It is traditional to use upper case letters in variable names, but we recommend using lower case letters fovariable names that serve internal purposes in the makefile, and reserving upper case for parameters that control implicit rules or for parameters that the user should override with command options (see [Overriding Variables](https://www.gnu.org/software/make/manual/html_node/Overriding.html)).

Makefile
pgiarrusso-sl: Pushed.

PR #57, opened 2025-12-11
Add support for generic diff errors in dune log

Janno:


rocq-tools/bin/code_quality_diff.ml
```diff
@@ -253,7 +259,7 @@ let analyse fmt ~before_dune ~after_dune ~before_globs ~after_globs =
         let (dangling_lines, w, e) = parse_lines (List.map parse_line output) in
         let dangling = List.map (fun (_, txt) -> dangling_output_warning src_file txt) dangling_lines in
         (List.rev_append w @@ List.rev_append dangling ws, List.rev_append e es)
-      | Cram ->
+      | Cram | Diff ->
```
Janno, line 262: We could get rid of the `Diff` constructor since we are not treating generic diff errors differently from cram test errors yet but we might want to do so in the future. I vote to keep it for the sake of clarity.

rlepigre-skylabs-ai (approved):

pgiarrusso-sl:
@rlepigre-skylabs-ai This should already be mergeable, right? Even if technically, we should confirm CI works.

rlepigre-skylabs-ai:
Yes, I think so. Re-running CI just in case (I think there is another PR with the same name, but that should not matter for this one, it can be merged before the other).

rocq-tools/bin/code_quality_diff.ml
```diff
@@ -182,6 +182,7 @@ let to_glob_out : strip_prefix:string -> string -> glob_out option =
 type dune_type =
```
pgiarrusso-sl, line 182: Can you explain (at least to yourself) why this stays here? I don't mind either way, but we should have an idea.

rocq-tools/bin/dune
```diff
@@ -114,10 +114,3 @@
  (modules perf_script)
  (package rocq-tools)
  (libraries rocq_tools yojson))
```
pgiarrusso-sl, line 116: Same question for all of `rocq-tools`. No need to have the answer in this set of PRs tho, there's already enough to integrate.

pgiarrusso-sl (approved):

rocq-tools/bin/code_quality_diff.ml
rlepigre-skylabs-ai: Good point. I guess that in my mind those are more generic. The one I moved is extremely specific to our setup.

rocq-tools/bin/dune
rlepigre-skylabs-ai: Could make sense to move everything indeed. I think they were initially put in `BRiCk` because that was the only public repo. We can change that later if we want to.

PR #60, opened 2025-12-12
Renaming directories with "bluerock" in their name

rlepigre-skylabs-ai:
This does not include renaming opam package names, we can do that in a second step that will require atomic PRs.


pgiarrusso-sl (approved):
LGTM!

PR #67, opened 2026-01-05
Rename [tc-db-info] into [rocq-tc-db-info].

rlepigre-skylabs-ai:
Also makes the plugin part of the same package.

Closing https://github.com/SkyLabsAI/workspace/issues/57.


rocq-tc-db-info/theories/dune
```diff
@@ -1,13 +1,6 @@
...
+ (package rocq-tc-db-info)
  (theories Ltac2)
- (plugins tc-db-info-plugin.plugin))
-
-; We generate a dummy file to force full dependency on the plugin package.
-; The contents of the file is not important (no caching issue here).
-(rule
- (targets dummy.v)
- (deps (package tc-db-info-plugin))
- (action (with-stdout-to %{targets} (echo "(* dummy file *)"))))
```
rlepigre-skylabs-ai, line 13: I have no idea what the point of that was, seems entirely useless. Maybe it was required because the plugin had its own package?

Janno: I don't remember either.

Janno (approved):

PR #68, opened 2026-01-12
Make dune file parser more robust

Janno:
Dune allows uninterpreted files called `dune` in conjunction with `data_only_dirs`. Our script can simply ignores files that do not follow the expected format.

### DIFF
--- a/scripts/coq_project_gen/gather-coq-paths.py
+++ b/scripts/coq_project_gen/gather-coq-paths.py
@@ -17,8 +17,11 @@ import re
 
 def find_key(symbol, obj):
     for x in obj:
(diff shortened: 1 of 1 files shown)


gmalecha-at-skylabs (approved):

gmalecha-at-skylabs:
This looks reasonable to me, but @rlepigre-skylabs-ai should probably confirm.

rlepigre-skylabs-ai (approved):

rlepigre-skylabs-ai:
It'd be better to actually interpret the `data_only_dirs` stanza, but it's probably not worth the trouble. We can try switching to generating `_RocqProject` files when the next version of VSRocq is released.

pgiarrusso-sl:
Merging!

Issue #69, opened 2026-01-07
Fix own_WeaklyObjective


pgiarrusso-sl:
Remaining TODOs:
- [x] new release — in https://github.com/SkyLabsAI/ci/pull/108, https://github.com/SkyLabsAI/ci/actions/runs/21307371704/job/61337736556
- [x] drop `own_WeaklyObjective`.

---

We admit `own_WeaklyObjective`, have `monPred_own_weakly_objective`, and the 2 don't match:
```
(* TODO upstream, fix [monPred_own_weakly_objective] with correct assumption. *)
#[global] Instance
  own_WeaklyObjective `{Σ : cpp_logic} {A : cmra} `{!HasOwn mpredI A} γ (a : A)  :
  WeaklyObjective (PROP := iPropI _) (own γ a).
Proof.
Set Printing All.
epose proof (monPred_own_weakly_objective γ a).
Unset Printing All.
Admitted.
```

The mismatch is `HasOwn` vs `inG` and is in the wrong direction — `HasOwn -> inG`, which can't be provable. So I _guess_ we're missing an `HasOwnWeaklyObjective` class — or a new member in existing classes. (It's unclear why we have 3 typeclasses in the first place).

hans89:
Your analysis is correct. `monPred_own_weakly_objective` depends on `inG` because `monPred_own` is defined as an embedding of iProp's own, which is defined using `inG`.
We can generalize it as follows. Assuming `monPred I PROP`,  and `HasOwn PROP A`. defined the monPred's `own` as the embedding of `PROP`'s `own`, then we will have `WeaklyObjective`.

```
Program Definition has_own_monpred_def  {PROP : bi} {A : cmra} `{!HasOwn PROP A} : HasOwn (monPredI I PROP) A := {|
    own := λ γ a , ⎡ own γ a ⎤%I
|}.
...

#[global] Instance monPred_own_weakly_objective  {PROP : bi} {A : cmra} `{!HasOwn PROP A} γ (x : A) :
    @WeaklyObjective I (monPredI I PROP) (own γ x).
```

pgiarrusso-sl:
Oooh, that could be nicer!

pgiarrusso-sl:
I need to sketch this, but it looks like we might have to replace `HasOwn mpredI A` assumptions with `HasOwn iPropI A` in some places. Possibly just where we need weak objectivity, which is just in brick-libcpp.

[...]

PR #70, opened 2026-01-21
Strengthen monPred_own_weakly_objective

pgiarrusso-sl:
## Downstream

https://github.com/SkyLabsAI/auto/pull/69
https://github.com/SkyLabsAI/bluerock.NOVA/pull/4
https://github.com/SkyLabsAI/bluerock.bhv/pull/10

## Description

With this MR, proving `WeaklyObjective (own (PROP := mpredI) γ a)` requires just `HasOwn iPropI A` instead of the stronger `inG Σ A`. We follow the plan in #69, and apply it to a few other definitions; discussed with @hans89 .

### Challenge: preserving termination of typeclass inference

Some new recursive instances can cause non-termination.
This MR adds recursive instances `HasOwn PROP A -> HasOwn (monPredI I PROP) A`, and these are unsafe since `HasOwn` lacks a mode: they can cause an infinite loop when applied to `HasOwn ?PROP A` by instantiating `?PROP := monPred ?I ?PROP'`. A `Hint Mode HasOwn + -` would fix this problem but require quite a few annotations. 

Janno suggested turning these instances into `Hint Extern` to perform mode-checking only for these instances.

### DIFF
--- a/rocq-skylabs-iris/theories/base_logic/monpred_own.v
+++ b/rocq-skylabs-iris/theories/base_logic/monpred_own.v
@@ -37,7 +37,7 @@ Section with_PROP.
     Next Obligation. solve_proper. Qed.
     Next Obligation. solve_proper. Qed.
     #[local] Definition has_own_monpred_aux : seal (@has_own_monpred_def). Proof. by eexists. Qed.
-    #[global] Instance has_own_monpred : HasOwn monPredI A := has_own_monpred_aux.(unseal).
+    #[local] Instance has_own_monpred : HasOwn monPredI A := has_own_monpred_aux.(unseal).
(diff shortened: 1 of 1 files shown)


pgiarrusso-sl:
Requesting review from @hans89 and @gmalecha-at-skylabs .

CI status: This already passed full build, but I'll confirm that it solves the original problem.

hans89:
LGTM

gmalecha-at-skylabs (approved):
Approving based on @hans89's review.

pgiarrusso-sl:
https://github.com/SkyLabsAI/ci/actions/runs/21255793902/job/61217300620#logs is a CI bug 💀

PR #73, opened 2026-02-05
Add semantics of the [move] builtin.

gmalecha-at-skylabs:
In some compilers (including Clang), `move` is a builtin.


pgiarrusso-sl (approved):
I have follow-ups but LGTM, merging!

PR #75, opened 2026-02-09
Enable generation of compile_commands.json.

gmalecha-at-skylabs:
This improves IDE support.


jhaag-skylabs-ai (approved):
This seems like the right way to enable generation of `compile_commands.json` for the `cpp2v` binary.

PR #87, opened 2026-03-03
[rocq-perf] Fix fork bomb when Rocq is not in the workspace.

### LINKED ISSUE #86: Performance shims invoke themselves when Rocq is not in the workspace
The change in https://github.com/SkyLabsAI/BRiCk/pull/85 is the source of the issue, before it `rocq` could not be located.

rlepigre-skylabs-ai:
Fixes https://github.com/SkyLabsAI/BRiCk/issues/86

### DIFF
--- a/rocq-tools/bin/rocq_perf.ml
+++ b/rocq-tools/bin/rocq_perf.ml
@@ -37,16 +37,20 @@ type files = {
   stdout : string;
 }
 
-let rocq_bin =
-  match Sys.getenv_opt "DUNE_SOURCEROOT" with
-  | None       -> "rocq"
-  | Some(root) ->
-  let rocq_bin = Filename.concat root "_build/install/default/bin/rocq" in
-  if Sys.file_exists rocq_bin then rocq_bin else "rocq"
+let rocq_command : args:string list -> string =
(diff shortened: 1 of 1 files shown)


rocq-tools/bin/rocq_perf.ml
```diff
@@ -37,16 +37,20 @@ type files = {
...
+let rocq_command : args:string list -> string =
+  let (cmd, leading_args) =
+    match Sys.getenv_opt "DUNE_SOURCEROOT" with
+    | None       -> ("rocq", [])
+    | Some(root) ->
+    let rocq_bin = Filename.concat root "_build/install/default/bin/rocq" in
+    if Sys.file_exists rocq_bin then
+      (rocq_bin, [])
+    else
+      ("opam", ["exec"; "--"; "rocq"])
```
rlepigre-skylabs-ai, line 49: This reinstates the hack that we were using before, in the case where Rocq is not in the workspace.

rlepigre-skylabs-ai: An alternative would be to pop the `PATH`, but not sure this would be robust.

PR #88, opened 2026-03-04
Fix constr/type confusion in `make_evar`

Janno:


Janno:
Linked to another branch but should be independent.

PR #89, opened 2026-03-04
Add test case for fixed bug in make_evar_in_level_env

Janno:


gmalecha-at-skylabs (approved):
A test case would be great here!

Janno:
Hehe. Good call. This PR is actually unfixing the fix that was already merged into master. I'll turn it into a PR that adds a test case.

PR #90, opened 2026-03-15
upstream various basic ltac2 functions

simon-skylabs:
Upstream a number of convenience Ltac2 functions about:
* `string`
* `option`
* `ident`
* `list`
* `fset`/`fmap`
* `binder`
* `reference`
* `message`
* `pp`
* `comparison`

As well as three novel functionalities:
* `Control.fold_left_goal_constrs` (and others) to traverse all the terms in a goal regardless where they appear;
* `Std.rename_ssr_vars` to rename all variables using the ssreflect naming scheme in context to remove the underscores and avoid warnings / errors;
* `List.bisect_partition` to partition a list using a predicate between the elements that satisfy the predicates and those that don't. It is built to minimize the number of calls to the predicate `p : 'a list -> bool` when testing multiple elements at once is less than linearly more expensive than testing a single element. It yields better performances than the straightforward `partition` when relatively few elements in the list *do not* satisfy the predicate. (see comment in code for more information)

### DIFF
--- a/ltac2-extra/theories/internal/compare.v
+++ b/ltac2-extra/theories/internal/compare.v
@@ -1,0 +1,17 @@
+(*
+ * Copyright (C) 2026 Skylabs AI, Inc.
+ *
+ * This software is distributed under the terms of the BedRock Open-Source
+ * License. See the LICENSE-BedRock file in the repository root for details.
+ *)
+
+Require Import skylabs.ltac2.extra.internal.init.
+
+(** Minor extensions to [Ltac2.String] *)
+Module Compare.
+  Import Ltac2.
+
+  Ltac2 compare_on (f : 'a -> 'b) (cmp : 'b -> 'b -> int) : 'a -> 'a -> int :=
+    fun x y => cmp (f x) (f y).
+
+End Compare.
--- a/ltac2-extra/theories/internal/compare.v
+++ b/ltac2-extra/theories/internal/compare.v
@@ -73,7 +73,7 @@ Module Comparison.
 
 End Comparison.
 
-(** Minor extensions to [Ltac2.String] *)
+(** Utilities for comparison functions *)
 Module Compare.
   Import Ltac2.
 
--- a/ltac2-extra/theories/internal/compare.v
+++ b/ltac2-extra/theories/internal/compare.v
@@ -7,6 +7,72 @@
 
 Require Import skylabs.ltac2.extra.internal.init.
 
+(** Functions to express comparisons in terms of a data type instead of integers. *)
+Module Comparison.
+  Import Ltac2.
+  Import Init.
+
+  Ltac2 Type 'a compare := 'a -> 'a -> comparison.
+
+  Ltac2 of_int (i : int) : comparison :=
+    if Int.lt i 0 then Lt
+    else if Int.gt i 0 then Gt
+    else Eq.
+
+  Ltac2 to_int (c : comparison) :=
+    match c with
+    | Lt => -1
+    | Eq => 0
+    | Gt => 1
+    end.
+
+  Ltac2 lift (cmp : 'a -> 'a -> int) : 'a -> 'a -> comparison :=
+    fun x y => of_int (cmp x y).
+
+  Ltac2 lower (cmp : 'a compare) : 'a -> 'a -> int :=
+    fun x y => to_int (cmp x y).
+
+  Ltac2 compare_on (f : 'a -> 'b) (cmp : 'b compare) : 'a -> 'a -> comparison :=
+    fun x y => cmp (f x) (f y).
+
+  Ltac2 lt (cmp : 'a compare) (x : 'a) (y : 'a) : bool :=
+    match cmp x y with
+    | Lt => true
+    | _ => false
+    end.
+  Ltac2 eq (cmp : 'a compare) (x : 'a) (y : 'a) : bool :=
+    match cmp x y with
+    | Eq => true
+    | _ => false
+    end.
+  Ltac2 gt (cmp : 'a compare) (x : 'a) (y : 'a) : bool :=
+    match cmp x y with
+    | Gt => true
+    | _ => false
+    end.
+  Ltac2 le (cmp : 'a compare) (x : 'a) (y : 'a) : bool :=
+    Bool.neg (gt cmp x y).
+  Ltac2 ge (cmp : 'a compare) (x : 'a) (y : 'a) : bool :=
+    Bool.neg (lt cmp x y).
+
+  Ltac2 rec lexicographical (cmp : (unit -> comparison) list) : comparison :=
+    match cmp with
+    | [] => Eq
+    | c :: cmps =>
+        let c := c () in
+        match c with
+        | Eq => lexicographical cmps
+        | _ => c
+        end
+    end .
+
+  Ltac2 lex2 (cmp_a : 'a compare) (cmp_b : 'b compare) : ('a * 'b) compare :=
+    fun (x0,x1) (y0,y1) => lexicographical [(fun () => cmp_a x0 y0);(fun () => cmp_b x1 y1)].
+  Ltac2 lex3 (cmp_a : 'a compare) (cmp_b : 'b compare) (cmp_c : 'c compare) : ('a * 'b * 'c) compare :=
+    fun (x0,x1,x2) (y0,y1,y2) => lexicographical [(fun () => cmp_a x0 y0);(fun () => cmp_b x1 y1);(fun () => cmp_c x2 y2)].
+
+End Comparison.
+
 (** Minor extensions to [Ltac2.String] *)
 Module Compare.
   Import Ltac2.
--- a/ltac2-extra/theories/internal/constr.v
+++ b/ltac2-extra/theories/internal/constr.v
@@ -481,15 +481,19 @@ Module Constr.
       Ltac2 invalid_arg (whence : string) (t : constr) : 'a :=
         invalid_arg' whence (Message.of_constr t).
 
+      Ltac2 of_app : constr -> (constr * constr array) := fun t =>
+        let rec go trm acc :=
+          match Unsafe.kind trm with
+          | Unsafe.App fn args => go fn (args :: acc)
+          | _ => (trm, Array.concat acc)
+          end in
+        go t [].
+
       Ltac2 of_ind_app_opt : constr -> (constr * constr array) option := fun t =>
-        match kind t with
-        | Ind _ _    => Some (t, Array.make 0 t)
-        | App c args =>
-            match kind c with
-            | Ind _ _ => Some (c, args)
-            | _       => None
-            end
-        | _          => None
+        let (c, args) := of_app t in
+        match kind c with
+        | Ind _ _ => Some (c, args)
+        | _       => None
         end.
 
       Ltac2 of_ind_app : constr -> constr * constr array := fun t =>
@@ -501,14 +505,10 @@ Module Constr.
 
       Ltac2 of_constructor_app_opt :
           constr -> (constr * instance * constr array) option := fun t =>
-        match kind t with
-        | Constructor _ inst => Some (t, inst, Array.make 0 t)
-        | App c args         =>
-            match kind c with
-            | Constructor _ inst => Some (c, inst, args)
-            | _       => None
-            end
-        | _          => None
+        let (c, args) := of_app t in
+        match kind c with
+        | Constructor _ inst => Some (c, inst, args)
+        | _       => None
         end.
 
       Ltac2 of_constructor_app : constr -> constr * instance * constr array :=
--- a/ltac2-extra/theories/internal/constr.v
+++ b/ltac2-extra/theories/internal/constr.v
@@ -68,6 +68,7 @@ Module Constr.
 
   Module Binder.
     Export Ltac2.Constr.Binder.
+    Import Ltac2.Bool.
 
     Ltac2 deconstruct (b : binder) : name * relevance * type :=
       (name b, relevance b, type b).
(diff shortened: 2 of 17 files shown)


ltac2-extra/theories/internal/control.v
```diff
@@ -73,6 +73,39 @@ Module Control.
...
+    match x with
+    | Hyp h => fprintf "hypothesis %I" h
+    | DefValue h => fprintf "local definition %I (value)" h
+    | DefType h => fprintf "local definition %I (type)" h
+    | Concl => fprintf "conclusion"
+    end.
+
+  (** List all the terms in the current goal including the conclusion and the value of local
+      definitions. The hypotheses are local definitions are listed in the order in which they appear
+      and the concludsion is listed last *)
```
gmalecha-at-skylabs:
```suggestion
      and the conclusion is listed last *)
```

ltac2-extra/theories/internal/fmap.v
```diff
@@ -0,0 +1,46 @@
+(*
+ * Copyright (C) 2022-2024 BlueRock Security, Inc.
```
gmalecha-at-skylabs:
```suggestion
 * Copyright (C) 2026 SkyLabs AI, Inc.
```

ltac2-extra/theories/internal/compare.v
```diff
@@ -0,0 +1,83 @@
...
+    end .
+
+  Ltac2 lex2 (cmp_a : 'a compare) (cmp_b : 'b compare) : ('a * 'b) compare :=
+    fun (x0,x1) (y0,y1) => lexicographical [(fun () => cmp_a x0 y0);(fun () => cmp_b x1 y1)].
+  Ltac2 lex3 (cmp_a : 'a compare) (cmp_b : 'b compare) (cmp_c : 'c compare) : ('a * 'b * 'c) compare :=
+    fun (x0,x1,x2) (y0,y1,y2) => lexicographical [(fun () => cmp_a x0 y0);(fun () => cmp_b x1 y1);(fun () => cmp_c x2 y2)].
+
+End Comparison.
+
+(** Minor extensions to [Ltac2.String] *)
```
rlepigre-skylabs-ai: I guess a copy-paste mistake?

ltac2-extra/theories/internal/fmap.v
```diff
@@ -0,0 +1,46 @@
...
+(** Minor extensions to [Ltac2.FMap] *)
+Module FMap.
+  Import Ltac2 Init Printf.
+  Export Ltac2.FMap .
+
+  Ltac2 alter (v0 : 'v) (k : 'k) (f : 'v -> 'v) (m : ('k,'v) FMap.t) : ('k,'v) FMap.t :=
+    match FMap.find_opt k m with
+    | Some v => FMap.add k (f v) m
+    | None => FMap.add k (f v0) m
+    end.
```
rlepigre-skylabs-ai, line 21: I wonder if we should rather use the usual "update" function of OCaml, which we could then simply expose as an optimization.

This would be something like:
```
Ltac2 update (k : 'k) (f : 'v option -> 'v option) (m : ('k, 'v) FMap.t) : ('k, 'v) FMap.t :=
  match f (FMap.find_opt k m) with
  | None => m
  | Some v => FMap.add k v m
  end.
```

ltac2-extra/theories/internal/fmap.v
```diff
@@ -0,0 +1,46 @@
...
+      (fun k v acc =>
+         if Option.is_none v then FMap.remove k acc
+      else acc ) map map in
+    FMap.mapi (fun _ v => Option.get v) map.
+
+  Ltac2 filteri (f : 'k -> 'a -> bool) : ('k, 'a) FMap.t -> ('k, 'a) FMap.t :=
+    filter_mapi
+      (fun k v => if f k v then Some v else None).
+
+  Ltac2 pp_fmap (pp_k : 'k pp) (pp_a : 'a pp) : ('k, 'a) FMap.t pp :=
```
rlepigre-skylabs-ai: Maybe this should be called just `pp`, and invoked qualified with `FMap.pp`? I usually prefer to go with as short names as possible, and disambiguate with module, but I don't have a very strong opinion.

ltac2-extra/theories/internal/fset.v
```diff
@@ -22,4 +24,12 @@ Module FSet.
     Ltac2 @ external evar_tag : evar FSet.Tags.tag :=
       "rocq-runtime.plugins.ltac2" "fmap_evar_tag".
   End Tags.
+
+  Ltac2 of_list (tag : 'a FSet.Tags.tag) (xs : 'a list) : 'a FSet.t :=
+    List.foldl FSet.add xs (FSet.empty tag).
+
+  Ltac2 pp_fset (pp_a : 'a pp) : 'a FSet.t pp :=
```
rlepigre-skylabs-ai: Maybe just `pp`?

ltac2-extra/theories/internal/init.v
```diff
@@ -7,6 +7,10 @@
 
 Require Import skylabs.ltac2.extra.internal.plugin.
 
+(** Avoids getting module [Corelib.Classes.Init] to shadow [internal.init.Init] when imported after
+    [internal.init]. *)
+Require Stdlib.Strings.String.
```
rlepigre-skylabs-ai, line 12: I don't quite understand the comment here, can you explain what this is about?

ltac2-extra/theories/internal/list.v
```diff
@@ -12,6 +12,16 @@ Module List.
...
 
+  Ltac2 rec tails (xs : 'a list) : 'a list list :=
+    match xs with
+    | [] => []
+    | _x :: xs =>
+        xs :: tails xs
+    end .
+
+  Ltac2 inits (xs : 'a list) : 'a list list :=
+    List.map List.rev (tails (List.rev xs)).
```
rlepigre-skylabs-ai: These seem like rather odd functions to me, what are they used for?

ltac2-extra/theories/internal/list.v
```diff
@@ -44,6 +54,77 @@ Module List.
       (xs : 'a list) (ys : 'b list) (acc : 'c) : 'c :=
     List.fold_right2 f xs ys acc.
 
+  Ltac2 rec first_some f xs :=
```
rlepigre-skylabs-ai: I would call this function `find_opt`, to match OCaml.

ltac2-extra/theories/internal/list.v
```diff
@@ -44,6 +54,77 @@ Module List.
...
+        match xs with
+        | [] => (List.rev acc, [])
+        | x :: xs =>
+
+            go (Int.sub n 1) xs (x :: acc)
+        end in
+    go n xs [].
+
+  (* [bisect_partition ps xs]
+     Assuming for some predicate [ps : 'a -> bool], [ps xs = List.for_all p xs], [bisect_partition] is
```
rlepigre-skylabs-ai, line 70:
```suggestion
     Assuming for some predicate [p : 'a -> bool], [ps xs = List.for_all p xs], [bisect_partition] is
```

rlepigre-skylabs-ai: Do you have an actual example of "batch predicate" `ps` for which you actually gain anything? I'm curious because I would intuitively think that the implementation of `ps` would likely need to do what you save by limiting calls to it.

ltac2-extra/theories/internal/string.v
```diff
@@ -48,4 +49,38 @@ Module String.
 
   Ltac2 newline () : string :=
     String.make 1 (Char.of_int 10).
+
+  Ltac2 is_prefix_of (pre : string) (s : string) : bool :=
```
rlepigre-skylabs-ai: I would call it `starts_with`, and the following function `ends_with` to follow OCaml, but no strong opinion here.

rlepigre-skylabs-ai (approved):
Looks good to me.

ltac2-extra/theories/internal/init.v
simon-skylabs: Yes, there's some weird with imports. When we write this in `internal/string.v`:

```
Require Import skylabs.ltac2.extra.internal.init.
Require Import skylabs.ltac2.extra.internal.constr.
Require Import skylabs.ltac2.extra.internal.char.

Require Import Stdlib.Strings.String.
```

the modules that now import `string.v`, when they try to import `Corelib.Classes.Init` instead of `internal.init.Init`, and so do other modules downstream and confusion ensues when referring to definitions such as `name`.

I have tried changing the ways `string.v` is imported but I'd rather try to not have a wrong way of importing `string.v`. What I find even more perplexing is that writing this in `string.v`:

```
Require Import Stdlib.Strings.String.

Require Import skylabs.ltac2.extra.internal.init.
Require Import skylabs.ltac2.extra.internal.constr.
Require Import skylabs.ltac2.extra.internal.char.
```

i.e. import `Stdlib.Strings.String` first does not address the problem.

What are your thoughts?

ltac2-extra/theories/internal/list.v
simon-skylabs: They are part of the standard Haskell library. But now that I think on it, they may make more sense in the context of lazy evaluation. 

In `Module Printf`, I used `tails` to implement a pretty printer for references. I provide one pretty printer, `pp_reference` that just prints the full name but the second one `pp_short_ref` prints the shortest unambiguous qualification of a name.

I can inline `tails` into `pp_short_ref` if you don't think it's a useful function to keep on its own.

ltac2-extra/theories/internal/list.v
simon-skylabs: `find_opt` already exists. Let me show you the signature of each side by side:

```
Ltac2 rec find_opt (f : 'a -> bool) (xs : 'a list) : 'a option :=
Ltac2 rec first_some (f : 'a -> 'b option) (xs : 'a list) : 'b option :=
```

I also think `first_some` doesn't sound good. What do you think of `find_map_opt`?

ltac2-extra/theories/internal/list.v
simon-skylabs: I know what you mean. I went from a tactic that took 5s to something under a second by using this. The thing is that `ps` has to be very expensive and that the price be related to something other than the size of the list.

In my situation, I have a tactic that takes a tactic as an argument and tries to extract a lemma statement such that the given tactic is capable of closing the goal. One part of this task is looking at a long list of intutionistic premises, most of which are not needed to finish the proof and see which one I can clear and still be able to finish. The thing to minimize here is the number of runs of the user tactic. It is likely that I can ask "can I clear everything" and the answer will be yes. But if one or two of the premises are needed, this can find which ones with a logarithmic number of calls.

ltac2-extra/theories/internal/init.v
rlepigre-skylabs-ai: Oh, it's because we do something like `Import Init.`?

ltac2-extra/theories/internal/list.v
rlepigre-skylabs-ai: I don't mind having them, but you're probably right they make more sense in Haskell.

ltac2-extra/theories/internal/list.v
rlepigre-skylabs-ai: Ah sorry, I missed this. Adding the type could be useful. :smile: 

I agree `find_map_opt` is probably a better name.

ltac2-extra/theories/internal/list.v
rlepigre-skylabs-ai: I see, makes sense. Thanks for the explanation!

ltac2-extra/theories/internal/fmap.v
Janno: Isn't there a problem with the fact that all Ltac2 closures are impure functions in the proofview monad? I don't know how to turn an Ltac2 function of type `'v option -> 'v option` into an equivalent OCaml value of the same type.

ltac2-extra/theories/internal/list.v
Janno: Ltac2 has support for lazy values so why not use that to encode the original intent? https://rocq-prover.org/doc/V9.1.0/corelib/Ltac2.Lazy.html

ltac2-extra/theories/internal/init.v
simon-skylabs: Yes that's right. I'm not completely clear on how `Require Stdlib.Strings.String` and `Import Init` interact in that occasion to be honest. Is it just that the most recently `Require`-ed file with a module `Init` is that one that `Import Init` picks?

ltac2-extra/theories/internal/list.v
simon-skylabs: I don't think it will see enough use to that. I'll just inline `tails` and we can revisit if it becomes an issue.

ltac2-extra/theories/internal/constr.v
```diff
@@ -79,6 +80,11 @@ Module Constr.
 
     Ltac2 map_type : (type -> type) -> binder -> binder := fun f b =>
       let (name, r, ty) := deconstruct b in unsafe_make name r (f ty).
+
+    Ltac2 equal (b0 : binder) (b1 : binder) : bool :=
+      Option.equal Ident.equal (Binder.name b0) (Binder.name b1) &&
+        Constr.equal (Binder.type b0) (Binder.type b1).
```
Janno, line 86: I think this is missing the comparison on `Binder.relevance`.

ltac2-extra/theories/internal/reference.v
```diff
@@ -0,0 +1,23 @@
...
+ * License. See the LICENSE-BedRock file in the repository root for details.
+ *)
+
+Require Import skylabs.ltac2.extra.internal.init.
+
+(** Some functions for [Std.reference]  *)
+Module Reference.
+  Import Ltac2 Init Std.
+
+  Ltac2 equal (r0 : Std.reference) (r1 : Std.reference) : bool :=
```
Janno, line 14:
```suggestion
  (* Remove this function once we bump to 9.2 which includes it *)
  Ltac2 equal (r0 : Std.reference) (r1 : Std.reference) : bool :=
```

ltac2-extra/theories/internal/constr.v
simon-skylabs: `relevance` is defined in terms of `relevance_var` and that one isn't defined in Ltac2 and doesn't have an API. Is it reasonable to think that all  `relevance_var` are equal or can we expose an equality test for it?

PR #94, opened 2026-03-24
ltac2-extra: evar and LevelEnv improvements.

Janno:


ltac2-extra/theories/internal/level_env.v
```diff
@@ -72,14 +72,18 @@ Module LevelEnv.
...
   Ltac2 level_subst' (env : t) : constr list := env.(rels).
-  Ltac2 level_subst (env : t) : constr -> constr :=
-    Constr.Unsafe.substnl (env.(rels)) 0.
+  Ltac2 level_subst (env : t) (c : constr) : constr :=
+    (* [substnl] assumes that we are instantiating binders and shifts [Rel]s
+    above [env.(length)] accordingly. This would make [level_subst] not
+    involutive. To fix this, we pre-emptively shift those binders by
+    [env.(length)]. The shift is then undone by [substnl]. *)
+    let n := env.(length) in
+    let c := Constr.Unsafe.liftn n (Int.add n 1) c in
```
Janno: This is really the only controversial part of this PR. No current client of this `level_subst` has `Rel`s that would be affected. So this `liftn` call is basically wasted time in 100% of the code in `main`. However, in my evar adventures, I have to juggle two contexts (the current goal's context and the evar's own context) and these two contexts may use the same name for different variables. The best way I could find to work with terms that use variables from both these contexts is to always treat one of them as a `Rel`ative context (and the other as a named context).

However, at some point I need to compute a closed term (in the original context). Since they have overlap in binder names, my only choice is to do this by making both of them `Rel`ative temporarily. And that's how I end up with a term in a `LevelEnv.t` that has additional `Rel`s (above the length of the `LevelEnv`) on which I eventually need to call `level_subst`.

Instead of calling `liftn` in my client code, I thought it might make sense to give `level_subst` a stronger spec and make it involutive in all cases. But I am not sure this is the right way to go.

Janno: I ended up undoing this and going with a completely different approach. `LevelEnv` now has support for stacked contexts and that's what I am using instead.

ltac2-extra/theories/internal/level_env.v
```diff
@@ -53,11 +60,17 @@ Module LevelEnv.
     List.foldl f (env.(decls)) acc.
 
   (** The empty environment *)
-  Ltac2 empty : t := { decls := []; length := 0; rels := [] }.
+  Ltac2 empty : t := { decls := []; offset := 0; length := 0; rels := [] }.
+
+  Ltac2 empty_o (offset : int) : t :=
+    { empty with offset }.
```
rlepigre-skylabs-ai: This function's name is a bit cryptic, and it's simple enough that we could just inline it.

ltac2-extra/theories/internal/level_env.v
```diff
@@ -154,15 +241,29 @@ Module LevelEnv.
...
+  Ltac2 named_subst (env : t) : constr -> constr :=
+    let ls :=
+      let fn rd (free, ls) :=
+        let (free, id) := Fresh.for_rel_decl free rd in
+        (free, Constr.Unsafe.make_var id :: ls)
+      in
+      let (_, ls) := LevelEnv.foldr fn env (Fresh.Free.of_goal (), []) in
+      ls
+    in
+    Constr.Unsafe.substnl ls (env.(offset)).
```
rlepigre-skylabs-ai:
```suggestion
    let (_, ls) :=
      let fn rd (free, ls) :=
        let (free, id) := Fresh.for_rel_decl free rd in
        (free, Constr.Unsafe.make_var id :: ls)
      in
      LevelEnv.foldr fn env (Fresh.Free.of_goal (), [])
    in
    Constr.Unsafe.substnl ls (env.(offset)).
```

rlepigre-skylabs-ai (approved):

ltac2-extra/theories/internal/level_env.v
Janno: We don't currently make `LevelEnv.t` private but that's for debugging reasons (and because private types did not exist when the code was written by Dave). In spirit, its fields should not be accessible (let alone writable) by clients. I could rename the function into `empty_w_offset`.

PR #95, opened 2026-03-25
Ltac2 library for combining term builders and term parsers

simon-skylabs:


ltac2-extra/theories/internal/fmap.v
```diff
@@ -35,12 +34,4 @@ Module FMap.
     filter_mapi
       (fun k v => if f k v then Some v else None).
 
-  Ltac2 pp (pp_k : 'k pp) (pp_a : 'a pp) : ('k, 'a) FMap.t pp :=
```
rlepigre-skylabs-ai, line 38: Why move this? I like the convention of having a `pp` method in modules.

ltac2-extra/theories/internal/option.v
```diff
@@ -40,4 +39,43 @@ Module Option.
...
+      match x with
+      | Some x => Val x
+      | None => Err (e ())
+      end.
+
+    Module Export Notations.
+      Ltac2 Notation "_alt!" f(thunk(self)) := _alt f.
+    End Notations.
+
+  End Ap.
```
rlepigre-skylabs-ai, line 79: I'm not sure I understand the conventions here, why is everything prefixed with `_`?

ltac2-extra/theories/internal/option.v
```diff
@@ -40,4 +39,43 @@ Module Option.
     | None => []
     end.
 
+  Ltac2 of_cps (mx : ('a, 'r) cps) : 'a option :=
+    mx (fun x => Some x).
+
+  Ltac2 mret (x : 'a) : 'a option := Some x.
```
rlepigre-skylabs-ai, line 45: Why not add `mbind` too then?

rlepigre-skylabs-ai (commented):
I'm not sure I'm following why we need so much CPS and monadic code. I had not seen that code before the upstreaming, so @Janno can probably give a more useful review.

ltac2-extra/theories/internal/fmap.v
simon-skylabs: I was trying to decouple `Constr` and `Printf`. The new `Constr` code uses `FSet` and I think I tied myself into a cycle there. 

It might be that the mistake is having `Constr` depend on `FSet` however. I notice that there's next to no recursive code in `Constr` and no sophisticated algorithms. Maybe there's a better home for `instantiate_prod`. `Std` maybe? Or maybe a new file named `Unsafe`?

ltac2-extra/theories/internal/option.v
simon-skylabs: Because Ltac2 doesn't have interfaces, I'm trying to highlight that each `Ap` module does similar stuff. 

To your question
> I'm not sure I'm following why we need so much CPS and monadic code.

I keep rubbing up against the fact that we can't write functions that work for arbitrary monads. The observation underlying much of this PR (and that I haven't spelled out now that I think of it) is that we can write generically monadic functions by making them operate on `cps` and then we have lifting and lowering functions between the monads that we care about and cps.

One nice illustration of this idea is `List.map_accum_XX`:

```
  Ltac2 rec map_accum_cps (f : 's -> 'a -> ('s * 'b, 'r) cps) (s : 's) (xs : 'a list) : ('s * 'b list, 'r) cps :=
    match xs with
    | [] => Cps.mret (s, [])
    | x :: xs =>
        Cps.bind (f s x) (fun (s, y) =>
        Cps.bind (map_accum_cps f s xs) (fun (s, ys) =>
        Cps.mret (s, y :: ys)))
    end.
  Ltac2 map_accum_opt (f : 's -> 'a -> ('s * 'b) option) (s : 's) (xs : 'a list) : ('s * 'b list) option :=
    Option.of_cps (map_accum_cps (fun s x => Option.bind (f s x)) s xs).
  Ltac2 map_accum_result (f : 's -> 'a -> ('s * 'b) result) (s : 's) (xs : 'a list) : ('s * 'b list) result :=
    Result.of_cps (map_accum_cps (fun s x => Result.bind (f s x)) s xs).
```

In `auto`, in `rocq-skylabs-auto-cpp/theories/auto/diff.v`, there's a module `proofmode` to operate on IPM goals. I needed to map over all the terms in an IPM goal. In one case I wanted to apply a function of type `constr -> constr result` and in the other case I wanted to apply a function of type `constr -> constr option`. I got to write a single traversal function, `BiGoal.map_cps`, and specialize it for the `option`, `result` and `identity` monads.

ltac2-extra/theories/internal/option.v
simon-skylabs: The function `Option.bind` already exists. 

Grepping around in the Ltac2 code base, I realize that this is the only `bind` function that the basic Ltac2 libraries provide. We could standardize `mbind` with little confusion if we want to.

ltac2-extra/theories/internal/char.v
```diff
@@ -42,4 +42,25 @@ Module Char.
...
+    let (b0, n) := lsb n in
+    let (b1, n) := lsb n in
+    let (b2, n) := lsb n in
+    let (b3, n) := lsb n in
+    let (b4, n) := lsb n in
+    let (b5, n) := lsb n in
+    let (b6, n) := lsb n in
+    let (b7, n) := lsb n in
+    Control.assert_true (Int.equal n 0) ;
+    '(Ascii $b0 $b1 $b2 $b3 $b4 $b5 $b6 $b7).
```
Janno: Maybe construct this unsafely so that it works when no goal is in focus?

Janno (approved):

ltac2-extra/theories/internal/option.v
rlepigre-skylabs-ai: Ah OK, if `bind` exists then that's fine. It seems to be the commonly-used name in OCaml anyway.

ltac2-extra/theories/internal/option.v
simon-skylabs: I just talked about `_` with @Janno and we came up with an alternative. When using module `FooBar` applicatively, we can `Import FooBar (Ap)` and write `Ap.apply` instead of `_apply`.

rlepigre-skylabs-ai: I see, thanks for the explanation!

ltac2-extra/theories/internal/fmap.v
rlepigre-skylabs-ai: Were does the cycle come from through `fmap.v`? Is it because it itself relies on `constr.v`?

simon-skylabs:
@Janno @rlepigre-skylabs-ai if you can review the `auto` PR, I think we can merge this. https://github.com/SkyLabsAI/auto/pull/124

PR #96, opened 2026-03-26
Typeclass-based Ltac2 dispatch.

gmalecha-at-skylabs:
A sketch proposal for resolving an Ltac2 function using typeclasses.


rocq-ltac2-tc-dispatch/plugin/lookup.ml
```diff
@@ -0,0 +1,28 @@
...
+    (Names.Id.of_string id)
+  in None
+  (* try
+    let global_ref = Tac2env.locate_ltac qualid in
+    let func_val = Tac2env.interp_global qualid in
+    if func_val.gdata_type = (0, GTypArrow(_, _))
+    then Some (fun tt -> _)
+    else None
+  with
+    Not_found -> None *)
```
gmalecha-at-skylabs: I don't really know how to complete this.

rocq-ltac2-tc-dispatch/README.md
```diff
@@ -0,0 +1,7 @@
+Extension of the Ltac2 standard library
+=======================================
+
+To use the library, require via the following command:
+```coq
+Require Import skylabs.ltac2.extra.extra.
```
gmalecha-at-skylabs:
```suggestion
Require Import skylabs.ltac2.tc_dispatch.lookup.
```

rocq-ltac2-tc-dispatch/README.md
```diff
@@ -0,0 +1,7 @@
+Extension of the Ltac2 standard library
```
gmalecha-at-skylabs:
```suggestion
Typeclass-based Ltac2 Lookups
```

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,88 @@
...
+    | List.cons ?c ?cs => f c :: list_of_list_constr f cs
+    end.
+
+  Ltac2 goal_dispatch () :=
+    let g := Control.goal () in
+    let query := constr:(Ltac2Lookup $g) in
+    (* let _ := printf "query=%t" query in *)
+    let inst := Constr.Unsafe.make (TC.resolve None query) in
+    (* let _ := printf "inst=%t" inst in *)
+    let flags := {
```
SkySkimmer: why not use RedFlags.all?

rocq-ltac2-tc-dispatch/plugin/lookup.ml
```diff
@@ -0,0 +1,47 @@
+
+open Ltac2_plugin
+open Tac2ffi
+(* open Tac2val *)
+open Tac2externals
+
+let resolve_ltac2_safe (mp : string list) (id : string) : (unit -> unit Proofview.tactic) option =
+  let qualid = Libnames.make_qualid
+    (Names.DirPath.make (List.map Names.Id.of_string mp))
+    (Names.Id.of_string id)
```
SkySkimmer: Why not take a single string (containing `.`) and call qualid_of_string on it instead of taking string list + string and calling make_qualid?
That way conversion is simpler (no list_of_list_constr) and you don't have to remember which order the list needs to be in.

rocq-ltac2-tc-dispatch/plugin/lookup.ml
```diff
@@ -0,0 +1,47 @@
...
+  in
+  let is_unit_typ (t : int Tac2expr.glb_typexpr) : bool =
+    (* TODO: why does Init.unit have two different representations? *)
```
SkySkimmer: because you don't normalize it (cf what Tac2typing_env.kind does)

rocq-ltac2-tc-dispatch/plugin/lookup.ml
```diff
@@ -0,0 +1,47 @@
...
+
+    | _ -> false
+  in
+  let is_unit_to_unit ((n, t) : Tac2expr.type_scheme) : bool =
+    Int.equal n 0 &&
+    match t with
+    | Tac2expr.GTypArrow (t1, t2) -> is_unit_typ t1 && is_unit_typ t2
+    | _ -> false
+  in
+  match Tac2env.locate_ltac qualid with
```
SkySkimmer: I guess we don't expose the full_path based API but if we did would you want to use it?
The difference with locate is that locate will resolve short names, so for instance locate on `case` will succeed when Control is imported but the full path based API would need `Ltac2.Control.case` regardless of importing.

rocq-ltac2-tc-dispatch/plugin/lookup.ml
```diff
@@ -0,0 +1,47 @@
...
+  | Tac2expr.TacConstant kn ->
+    let data = Tac2env.interp_global kn in
+    if is_unit_to_unit data.Tac2env.gdata_type
+    then
+      let v = Tac2interp.eval_global kn in
+      Some (repr_to (fun1 unit unit) v)
+    else None
+  | Tac2expr.TacAlias _ ->
+    None
+  | exception _ -> None
```
SkySkimmer:
```suggestion
  | exception Not_found -> None
```

rocq-ltac2-tc-dispatch/plugin/lookup.ml
```diff
@@ -0,0 +1,47 @@
...
+
+
+let define s =
+  define Tac2expr.{ mltac_plugin = "ltac2_tc_dispatch"; mltac_tactic = s }
+
+let _ =
```
SkySkimmer:
```suggestion
let () =
```

rocq-ltac2-tc-dispatch/plugin/lookup.ml
```diff
@@ -0,0 +1,47 @@
...
+    | _ -> false
+  in
+  match Tac2env.locate_ltac qualid with
+  | Tac2expr.TacConstant kn ->
+    let data = Tac2env.interp_global kn in
+    if is_unit_to_unit data.Tac2env.gdata_type
+    then
+      let v = Tac2interp.eval_global kn in
+      Some (repr_to (fun1 unit unit) v)
+    else None
```
SkySkimmer: might want to return a result with some error message to distinguish between "could not resolve" "resolved to non constant (Abbreviation)" and "resolved to constant of incorrect type"

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,88 @@
...
+  Ltac2 Type exn ::= [ UnexpectedConstr (string * constr) ].
+
+  Ltac2 string_of_pstring_constr (c : constr) : string :=
+    match Constr.Unsafe.kind c with
+    | Constr.Unsafe.String s =>
+        (* If it's a known constant, we extract it *)
+        Pstring.to_string s
+    | _ =>
+        (* Fallback: direct conversion for literal strings *)
+        throw (UnexpectedConstr("Expected a pstring", c))
```
SkySkimmer: these comments feel kinda strange
how is throw a fallback?
what does "known" mean in "known constant"?

rocq-ltac2-tc-dispatch/plugin/lookup.ml
gmalecha-at-skylabs: This makes sense. Alternatively, I was thinking that we could get the module path from the constant of the type class instance. That makes it less fragile to move things around.

rocq-ltac2-tc-dispatch/plugin/lookup.ml
gmalecha-at-skylabs: For this use case, I would want to use definite references. Resolution at this level seems very awkward.

rocq-ltac2-tc-dispatch/plugin/lookup.ml
gmalecha-at-skylabs: I think this is a good idea, but I need to dig a bit more into the plugin API to figure out how to raise exceptions and such.

SkySkimmer: I said return result not raise though?

gmalecha-at-skylabs: I actually need to figure out how to do that as well. This API is pretty foreign to me.

rocq-ltac2-tc-dispatch/dune-project
```diff
@@ -0,0 +1,22 @@
...
+
+(generate_opam_files)
+
+(maintainers
+ "Gregory Malecha <gregory@skylabs-ai.com>")
+
+(authors
+ "Gregory Malecha <gregory@skylabs-ai.com>")
+
+(license MIT)
```
rlepigre-skylabs-ai: Why is there a `LICENSE-LGPL` file then?

rocq-ltac2-tc-dispatch/plugin/dune
```diff
@@ -0,0 +1,4 @@
+(library
+ (name ltac2_tc_dispatch)
+ (public_name ltac2-tc-dispatch.plugin)
+ (libraries coq-core.plugins.ltac2 rocq-extra stdlib-extra ltac2-extra.plugin))
```
rlepigre-skylabs-ai:
```suggestion
 (libraries rocq-runtime.plugins.ltac2 rocq-extra stdlib-extra ltac2-extra.plugin))
```

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,76 @@
...
+Require Import Stdlib.Strings.PrimString.
+Require Import skylabs.ltac2.extra.extra.
+
+Declare ML Module "ltac2-tc-dispatch.plugin".
+
+(* The typeclass: P is the goal/type, path is the module hierarchy, name is the tactic *)
+Class Ltac2Lookup (P : Prop) := {
+    ltac2_path : list string;
+    ltac2_name : string;
+  }.
```
rlepigre-skylabs-ai: I guess we could add a new vernacular command for populating this class, and also checking the type of the given Ltac2 value.

Janno:
I wonder if we should use dummy Rocq definitions to get the module paths automatically right. The simplest way would be to demand a definition of the same name as the tactic. That definition is used in the instance. No strings. At registration time, we check that a tactic of the same name exists. (And that it is not private.. we should check that anyway.)

This should allow us to not mention any module paths and Rocq will make sure that everything keeps working across module boundaries.

We could do the check in tactic-in-terms (or TC search) in the class constructor if we wanted to avoid a new vernacular command.

gmalecha-at-skylabs:
@Janno I mentioned a similar idea on the Rocq Zulip. I'm not completely certain how to implement this, but all the data should be there.

On a related note, I was thinking that we might be able to support:
1. Calling regular Ltac tactics
2. Passing arguments to tactics from the arguments of the typeclass, e.g. a `constr` for each argument.

Both seem optional, but could be useful.

rocq-ltac2-tc-dispatch/dune-project
gmalecha-at-skylabs: We should change to LGPL to encourage contributions over forking.

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,83 @@
...
```
gmalecha-at-skylabs: I think it would be useful to decompose the idea of dispatch from what happens when we get a match. This allows extending what happens. If we make the `Handler` type an opaque wrapper around `unit` (like it done in, e.g., Mtac), then we can allow more interpretations of this sort of thing.

```coq
Variant Handler (Goal : Type) : Type :=
| CallLtac2 (_ : unit) (* the value needs to be a constant *)
| Apply {into} (pf : into -> Goal). (* effectively this is a fast path for `apply pf` *)

Class Dispatch (Goal : Type) (h : Handler Goal) := {}.
```

I don't necessarily like the `Goal` index, but it could be okay.

One other thing, in order to implement something like `match goal`, you might want to support backtracing successes. For example, if you wrote something like:

```coq
Instance WithHyp (H : Assumption True) {Goal} : Dispatch (CallLtac2 MyClear H) Goal
```

Then, if there were multiple solutions to `Assumption`, then you would call the tactic on the second if the  tactic failed on the first.

Janno: I did not yet implement `Apply` (because I don't really understand how it is useful) but I addressed the other parts of your comment.

Janno:
I have made a lot of changes and added a new feature that lets us pass in `constr`s. I am not entirely sure how useful it will be but it seemed natural.

Obvious TODOs:
- Figure out if we want a vernacular command
- If we don't want a command, write a tactic-in-term notation wrapper for `CallLtac2` that makes sure the tactic exists and has the right type (at least at that time)
- Add a warning for tactics that could not be found at runtime. If we allow circumventing the definition-time check this could allow some kind of late binding that might be useful with certain module (type)/functor constructions.

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -12,63 +12,69 @@ Require Import skylabs.ltac2.extra.extra.
...
-              Message.concat [Message.of_string "Could not find: ";
-                              Message.of_string n_ltac2] in
-            Control.zero (Tactic_failure (Some err))
-        end
-    | _ =>
-        Control.zero (Tactic_failure (Some (Message.of_string "Could not reduce instance to record.")))
-    end.
-
+    let c :=
+      lazy_match! reduced_inst with
```
gmalecha-at-skylabs, line 92: I worry slightly that if we use `reduced_inst` here then we will be reducing the arguments which might be problematic. I could be wrong though.

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -12,63 +12,69 @@ Require Import skylabs.ltac2.extra.extra.
...
-  Ltac2 Type exn ::= [ UnexpectedConstr (string * constr) ].
 
-  Ltac2 string_of_pstring_constr (c : constr) : string :=
-    match Constr.Unsafe.kind c with
-    | Constr.Unsafe.String s => Pstring.to_string s
-    | _ => throw (UnexpectedConstr("Expected a pstring", c))
-    end.
+#[universes(polymorphic,cumulative)]
+Inductive Dispatch (goal : Type) (handler : Handler goal) := {}.
+Existing Class Dispatch.
```
gmalecha-at-skylabs: We should add some docs to both of these.

gmalecha-at-skylabs (commented):
@Janno the changes here look great! Thanks so much for taking them up.

rocq-ltac2-tc-dispatch/theories/lookup.v
gmalecha-at-skylabs: That's a good call!

rocq-ltac2-tc-dispatch/theories/lookup.v
Janno: You are right. I wanted to change it to weak head reduction but then I staged the changes differently and ended up forgetting about it. I'll fix it.

rocq-ltac2-tc-dispatch/dune-project
gmalecha-at-skylabs:
```suggestion
(maintainers
 "Jan-Oliver Kaiser <janno@skylabs-ai.com>"
  "Gregory Malecha <gregory@skylabs-ai.com>")

(authors
 "Gregory Malecha <gregory@skylabs-ai.com>"
 "Jan-Oliver Kaiser <janno@skylabs-ai.com>")

(license LGPL-2.1)
```

rocq-ltac2-tc-dispatch/theories/crush_ext.v
```diff
@@ -0,0 +1,22 @@
+Require Ltac2.Ltac2.
+Require Import skylabs.ltac2.tc_dispatch.lookup.
+
+(**
+[crush_ext] is An extensible version of [crush].
+
+Users can add new tactics / strategies by adding [Ltac2Lookup] hints
+to the [crush_ext] database.
```
gmalecha-at-skylabs, line 11:
```suggestion
[crush_ext] is An extensible version of [crush].

Users can add new tactics / strategies by adding [Dispatch] hints
to the [crush_ext] database.
```

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,175 @@
...
+(** Instances of [Dispatch goal handler] instruct the [goal_dispatch] tactic to
+    apply [handler] to [goal]. The resolution follows normal typeclass search
+    logic, i.e. priorities and opacity are respected. *)
+#[universes(polymorphic,cumulative)]
+Inductive Dispatch (goal : Type) (handler : Handler goal) := {}.
+Existing Class Dispatch.
+
+Ltac2 Log Flag tc_dispatch.
+
+Module ltac2.
```
gmalecha-at-skylabs, line 52: Maybe we should separate the tactics from the interface?

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,175 @@
...
+(** [CallLtac2 tac_ref] represents a strategy that calls the Ltac2 tactic
+    referenced by [tac_ref] on [goal]. *)
+| CallLtac2 (ltac2_tac : Ltac2Ref.t) : Handler goal.
+#[global] Arguments CallLtac2 {_} _.
+
+(** Instances of [Dispatch goal handler] instruct the [goal_dispatch] tactic to
+    apply [handler] to [goal]. The resolution follows normal typeclass search
+    logic, i.e. priorities and opacity are respected. *)
+#[universes(polymorphic,cumulative)]
+Inductive Dispatch (goal : Type) (handler : Handler goal) := {}.
```
gmalecha-at-skylabs: Since we don't need eliminators.

```suggestion
Variant Dispatch (goal : Type) (handler : Handler goal) := {}.
```

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,175 @@
...
+End Ltac2Ref.
+
+(** [Handler goal] represents strategies to apply to goals matching the type
+    [goal]. They are registered using [Dispatch] instances. *)
+#[universes(polymorphic,cumulative)]
+Inductive Handler (goal : Type) : Prop :=
+(** [CallLtac2 tac_ref] represents a strategy that calls the Ltac2 tactic
+    referenced by [tac_ref] on [goal]. *)
+| CallLtac2 (ltac2_tac : Ltac2Ref.t) : Handler goal.
+#[global] Arguments CallLtac2 {_} _.
```
gmalecha-at-skylabs, line 42:
```suggestion
#[universes(polymorphic,cumulative)]
Variant Handler (goal : Type) : Prop :=
(** [CallLtac2 tac_ref] represents a strategy that calls the Ltac2 tactic
    referenced by [tac_ref] on [goal]. *)
| CallLtac2 (ltac2_tac : Ltac2Ref.t) : Handler goal.
#[global] Arguments CallLtac2 {_} _.
```

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,175 @@
...
+      let (h, args) := Constr.decompose_app c in
+      (check_constant h, Array.to_list args)
+    in
+    match resolve_ltac2 r args with
+    | Some f => f ()
+    | None =>
+        let msg := fprintf "Could not find Ltac2 tactic: %t" c in
+        log[tc_dispatch, 1] "%m" msg;
+        Control.zero (Tactic_failure (Some msg))
+    end.
```
gmalecha-at-skylabs, line 110:
```suggestion
  (** [goal_dispatch_with dbs] resolves a [Dispatch] instance from the databases
        <dbs> and interprets the handler. This effectively acts like a <match goal>
        where the branches come from [Dispatch] hints in the the hint databases.
        
        The tactic runs the handlers in the order that they are found by typeclass search.
        The tactic completes successfully if any matching handler succeeds, otherwise
        it fails with `Control.zero`. If a handler raises and exception, the search stops
        and the exception is passed through.
        
        NOTE: [Dispatch] instances should not be directly added to <typeclass_instances>
        because <typeclass_instances> is used in all searches.
     *)
  Ltac2 goal_dispatch_with (dbs : ident list option) :=
    let g := Control.goal () in
    let query := open_constr:(Dispatch $g _) in
    log[tc_dispatch, 10] "query: %t" query;
    let _ := Constr.Unsafe.make (TC.resolve dbs query) in
    let inst := lazy_match! query with | Dispatch _ ?h => h end in
    log[tc_dispatch, 10] "inst: %t" inst;
    let flags := { RedFlags.all with Std.rStrength := Std.Head } in
    let reduced_inst := Std.eval_lazy flags inst in
    log[tc_dispatch, 10] "reduced_inst: %t" reduced_inst;

    let c :=
      lazy_match! reduced_inst with
      | CallLtac2 ?c => c
      | _ =>
          let msg := fprintf "Could not reduce %t to record. Stuck at: %t" inst reduced_inst in
          log[tc_dispatch, 1] "%m" msg;
          Control.zero (Tactic_failure (Some msg))
      end
    in
    let (r, args) :=
      let (h, args) := Constr.decompose_app c in
      (check_constant h, Array.to_list args)
    in
    match resolve_ltac2 r args with
    | Some f => f ()
    | None =>
        let msg := fprintf "Could not find Ltac2 tactic: %t" c in
        log[tc_dispatch, 1] "%m" msg;
        Control.zero (Tactic_failure (Some msg))
    end.
```

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,175 @@
...
+    let g := Control.goal () in
+    let deps := Constr.Vars.vars_really_needed g in
+    let f v :=
+      if Control.is_section_variable v then
+        let msg := fprintf "Signature %t depends on section variable %I" g v in
+        Control.zero (Invalid_argument (Some msg))
+      else
+        ()
+    in
+    List.iter f (FSet.elements deps).
```
gmalecha-at-skylabs: Is this correct?

```suggestion
  (** Checks that the current proof does not require any <Section> variables.
        Fails with <Control.zero> if any <Section> variable is used.
   *)
  Ltac2 check_requires_no_section_vars () :=
  (* We want to make sure that no section variables are used. There might be
     arguments already introduced. We revert all non-section variables and check
     that the resulting type does not depend on section variables. *)
    let hyps := Control.hyps () in
    let to_revert :=
      let f (i, _, _) := if Control.is_section_variable i then None else Some i in
      List.map_filter f hyps
    in
    Std.revert to_revert;
    let g := Control.goal () in
    let deps := Constr.Vars.vars_really_needed g in
    let f v :=
      if Control.is_section_variable v then
        let msg := fprintf "Signature %t depends on section variable %I" g v in
        Control.zero (Invalid_argument (Some msg))
      else
        ()
    in
    List.iter f (FSet.elements deps).
```

rocq-ltac2-tc-dispatch/theories/lookup.v
```diff
@@ -0,0 +1,175 @@
...
+          else
+            ()
+        in
+        List.iter f (FSet.elements vars);
+        Control.refine (fun _ => t)
+    | None =>
+        let msg := fprintf "Could not find Ltac2 tactic: %t" h in
+        log[tc_dispatch, 1] "%m" msg;
+        Control.zero (Invalid_argument (Some msg))
+    end.
```
gmalecha-at-skylabs:
```suggestion
  (** Checks whether the given term depends on any <Section> variables. *)
  Ltac2 check_ref_uses_section_vars (t : preterm) :=
    let t := Constr.pretype t in
    let (h, args) := Constr.decompose_app t in
    let r := check_constant h in
    match resolve_ltac2 r (Array.to_list args) with
    | Some _ =>
        let vars := Constr.Vars.vars_really_needed t in
        let f v :=
          if Control.is_section_variable v then
            let msg := fprintf "Ltac2Ref %t depends on section variable %I" h v in
            Control.zero (Invalid_argument (Some msg))
          else
            ()
        in
        List.iter f (FSet.elements vars);
        Control.refine (fun _ => t)
    | None =>
        let msg := fprintf "Could not find Ltac2 tactic: %t" h in
        log[tc_dispatch, 1] "%m" msg;
        Control.zero (Invalid_argument (Some msg))
    end.
```

Janno (approved):

rocq-ltac2-tc-dispatch/theories/lookup.v
Janno: They are not easy to disentangle because of the overlap in code required for `make_ltac2_ref`, `[ltac2 ..]` and the actual tactic. I think we can keep them together for now.

PR #97, opened 2026-03-27
ltac2 prelude tweaks

simon-skylabs:


Janno:
Hm, I hadn't realized `RelevanceVar` already landed. We cannot write an equality function for relevance. Fortunately, I just checked the OCaml code and `Constr.equal` ignores the name and relevance of `fun` and `forall` entirely. I think that means we do not need to worry about relevance at all.

However, this brings me to a question: why do we ever need to test binders for equality? Testing the names for equality makes sense. Testing the types for equality is already very, very suspect but could sometimes be useful. But both at the same time? I cannot imagine a use case.

Janno (approved):

PR #99, opened 2026-03-30
Feat/clang22 support

gmalecha-at-skylabs:
Closes #66

PR #103, opened 2026-03-31
Support for clang22.0.0

gmalecha-at-skylabs:
This release of clang seems to be effectively broken.


gmalecha-at-skylabs:
Supporting this pre-release version seems unnecessary.

pgiarrusso-sl:
Oh, the real support is in https://github.com/SkyLabsAI/BRiCk/pull/99.

PR #107, opened 2026-04-02
Upstreamed some definitions and lemmas from auto/lazy/big_sep/prelude.v

LennartATSkylabsAI:


rocq-skylabs-prelude/theories/list_numbers.v
```diff
@@ -1635,6 +1639,94 @@ Section listZ.
     - split => [|[]] //.
   Qed.
 
+  (** For every occurrences of [bool_decide] in the conclusion where the conclusion is known to be
+      true or known to be false, perform the substitution. *)
+  Ltac rw_bool_decide :=
```
simon-skylabs, line 1644: I think we can replace `rw_bool_decide` with `simpl_decide_with lia` which is defined a bit higher. I don't think we can replace `clear_decided_terms` but I think we should leave a note that there's a common abstraction that we should formulate.

simon-skylabs (approved):
Looks good! Thanks for doing the changes. After this last change, I think we can merge.

rocq-skylabs-prelude/theories/list_numbers.v
LennartATSkylabsAI: Done in next commit.

PR #110, opened 2026-04-08
Bug in name printing

gmalecha-at-skylabs:
The logic was re-printing the name itself.


jhaag-skylabs-ai (approved):
Approving because these changes seem reasonable based on my understanding of:
1. the clang AST
2. the BRiCk AST

I will admit that I haven't seen a manifestation of the underlying bug so I can't validate whether or not this fix is complete.

PR #114, opened 2026-04-14
Remove the initial state from [Sts]

gmalecha-at-skylabs:
Having the initial state is not very Markovian and it gets in the way. Operations that require the explicit state can take it separately.


pgiarrusso-sl (approved):

gmalecha-at-skylabs:
@simon-skylabs expressed some concern with this in a conversation, but I'm still thinking that this change would be a significant improvement.

PR #121, opened 2026-04-24
Fix and document backtracking in ltac2-tc-dispatch

Janno:


ltac2-extra/theories/internal/tc.v
```diff
@@ -14,6 +14,32 @@ Module TC.
...
+        ]
+      ];
+    tc.
+
+  (* [resolve_dbs dbs t] is a non-backtracking version of [resolve_all_dbs dbs t],
+     i.e. it commits to the first instance that solves the goal. *)
+  Ltac2 resolve_dbs dbs t :=
+    Control.once (fun () => resolve_dbs_all dbs t).
+
+  #[deprecated(since="20260424", note="Use [resolve_dbs] or [resolve_dbs_all]")]
```
gmalecha-at-skylabs, line 42: Since this isn't really published yet, it probably makes sense to avoid deprecated functions at this point. If we don't want people to use this, then we should just remove it.

gmalecha-at-skylabs (approved):

gmalecha-at-skylabs:
@Janno how do you feel about getting this merged?

PR #124, opened 2026-04-30
Script to fix transitive dependencies for dune

gmalecha-at-skylabs:
Coded by Codex.

### DIFF
--- a/scripts/coq_project_gen/README.md
+++ b/scripts/coq_project_gen/README.md
@@ -1,4 +1,4 @@
-# _CoqProject generator
+# _CoqProject generator and dune dependency sync
 
 This script will generate an approximate `_CoqProject` file from a dune project, by parsing
 all `rocq.theory` stanzas in `dune` files to generate `-Q` mappings.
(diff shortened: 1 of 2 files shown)


scripts/coq_project_gen/README.md
```diff
@@ -22,3 +22,56 @@ If desired, create a `your_dune_workspace/coq.flags` file containing a
...
+### Notes
+
+- The script rewrites only the `(theories ...)` subform and leaves the rest of
+  the file byte-identical.
+- Formatting inside the rewritten `theories` stanza is normalized to one
+  dependency per line.
+- With `--no-normalize`, the script preserves the existing dependency order and
+  only rewrites a file when newly discovered recursive dependencies must be
+  appended.
+- Comments inside the old `theories` stanza are not preserved.
```
pgiarrusso-sl, line 75: This description seems mostly reasonable.

I have a concern about normalization — I'm 90% sure we don't need to act on it, but I struggle to avoid mentioning it.

----

But reordering deps and dropping comments would hinder undoing this change in the far future.

1. Maybe we'd just write a script that finds minimal dependencies by graph traversal (find the smallest set whose closure is the same). It's not what I'd want but it's good enough.
2. Why's that not perfect? If I depend directly on A and B, it is more robust to depend on both. But if A depends on B, suggestion 1 would instead drop the dependency on B. If then A stops depending on B, our code would break.

But to be honest, the failure is rare and easy enough to fix, so maybe "do nothing now" is enough.

pgiarrusso-sl (approved):

scripts/coq_project_gen/sync-rocq-theory-deps.py
```diff
@@ -0,0 +1,507 @@
+#!/usr/bin/env -S uv run --script
```
pgiarrusso-sl, line 1: I have not reviewed this agent-generated code, and it comes with no tests.

1. Either ask Codex to add tests, _or_
2. Block this PR until we can review the generated changes.

Please don't resolve until either solution is done.

pgiarrusso-sl (changes requested):
LGTM up to the lack of tests, but I'll approve once that's fixed.

scripts/coq_project_gen/README.md
gmalecha-at-skylabs: The current algorithm looks at all of the dependencies explicitly listed, and then adds in any transitive dependencies that are missing. We could try to set it up so that the direct dependencies are explicitly separated, that way we end up with something like:

```
(theories
   dep1 dep2
   ;; transitive dependencies, regenerate if necessary
   tdep1 tdep2 tdep3)
```

Would that be better?

pgiarrusso-sl: Yes, that'd be great. And such files would be easier to review!

gmalecha-at-skylabs:
I propose that to harden this, we port it to Ocaml and test it there. Similar to what is done in #129.

pgiarrusso-sl:
> I propose that to harden this, we port it to Ocaml and test it there. Similar to what is done in https://github.com/SkyLabsAI/BRiCk/pull/129.

Up to you.

gmalecha-at-skylabs:
Subsumed by #130

PR #129, opened 2026-05-01
Port _CoqProject generation to dune

gmalecha-at-skylabs:
This makes it easier to test and avoids the python dependency.


rocq-dune-tools/README.md
```diff
@@ -0,0 +1,38 @@
+Rocq Dune Tools
+===============
+
+This project packages a small OCaml executable, `dune-rocqproject`, for
+generating `_CoqProject` content from a Rocq/Dune workspace.
+
+It replaces the behavior of:
+
+- `scripts/coq_project_gen/gather-coq-paths.py`
+- `scripts/coq_project_gen/gen-_CoqProject-dune.sh`
```
gmalecha-at-skylabs: It looks like this file needs to be fixed

```suggestion
```

pgiarrusso-sl: I thought that's correct? But we'd have to also replace those tools?

gmalecha-at-skylabs: The idea is simply to not expose the historical artifacts.

gmalecha-at-skylabs:
@rlepigre-skylabs-ai suggests that this can now be done in the current version of `dune`. If so, we can drop this functionality and pivot to that when it is convenient.

gmalecha-at-skylabs:
Better to do in dune.

PR #130, opened 2026-05-05
Add script to fix transitive dependencies for dune

gmalecha-at-skylabs:
Port of #124 to ocaml


rocq-dune-tools/lib/parse.ml
```diff
@@ -0,0 +1,238 @@
...
+    if i >= length then failf "Unterminated string literal"
+    else if Char.equal text.[i] '\\' then loop (i + 2)
+    else if Char.equal text.[i] '"' then i + 1
+    else loop (i + 1)
+  in
+  loop (index + 1)
+
+(* Sexplib gives us structure, but not source locations. This small scanner is
+   only for finding list spans so we can rewrite just the (theories ...) form
+   without reserializing the entire dune file. *)
```
rlepigre-skylabs-ai: There is [this module](https://github.com/janestreet/sexplib/blob/481e4a51196bd76903ca1f7e116af2387a7e7d87/src/sexp_intf.ml#L135-L136) that seems to provide annotated s-expression AST, so this seems whole infrastructure is probably not necessary.

rocq-dune-tools/lib/rewrite.ml
```diff
@@ -0,0 +1,183 @@
...
+
+let read_process_output ic =
+  let buffer = Buffer.create 256 in
+  ( try
+      while true do
+        Buffer.add_string buffer (input_line ic) ;
+        Buffer.add_char buffer '\n'
+      done
+    with End_of_file -> () ) ;
+  Buffer.contents buffer
```
rlepigre-skylabs-ai: This is basically `In_channel.input_all`.

rocq-dune-tools/lib/rewrite.ml
```diff
@@ -0,0 +1,183 @@
...
+
+let emit_diff ~display old_text new_text =
```
rlepigre-skylabs-ai: This tool does not need to emit diffs IMO: everything is in git, so I would just run it in a clean state and check the result with `git diff` / `make peek` in the workspace to get the same effect. Did you specifically request this feature, or did codex just add it?

rocq-dune-tools/lib/workspace.ml
```diff
@@ -0,0 +1,91 @@
...
+      workspace
+  | None -> (
+    match find_workspace_root_by_walk cwd with
+    | Some root ->
+        {cwd; root}
+    | None ->
+        failf
+          "failed to run dune describe workspace, and no dune-workspace file \
+           was found in %s or any parent directory"
+          (Fpath.to_string cwd) )
```
rlepigre-skylabs-ai: It's a bit odd to try two things here, I believe that the strategy we use in other places is probably more robust anyway (getting the value of `DUNE_SOURCE_ROOT` in the output of `dune exec -- env`, or something like that).

rocq-dune-tools/lib/graph.ml
```diff
@@ -0,0 +1,218 @@
+open Support
+
+type t =
+  { theory_index: theory_stanza list StringMap.t
+  ; closure_cache: (string, StringSet.t) Hashtbl.t }
```
rlepigre-skylabs-ai: I would use `ocamlgraph` (see [here](https://backtracking.github.io/ocamlgraph/ocamlgraph/Graph/Pack/Digraph/index.html)) or some other widely used library to compute the transitive closure.

rlepigre-skylabs-ai (commented):
At a first glance, this looks like this suffers from the same kind of issues as other LLM-generated code I've seen (for OCaml or for Python). Mainly, it seems to write way too much code, and to do certain things in circumvented ways, which makes it hard to review properly.

It also seems to make up issues, like missing source code positions in `sexplib`, so it works around a non-existant problem which leads to pretty sketchy lexing / parsing code. Not sure if this is due to lack of training on code using that library, or if it's just a form of hallucination.

That being said, we might be able to iterate using more prompting, pointing codex at the right `sexplib` interface for example.

Another thought I have is that it might be worth having codex write interfaces in `.mli` files first, before writing the code. It might be a good way to have it take smaller steps, which might give better results. This would also give us a chance to check that it's using the right types (e.g., an existing graph implementation, or s-expressions with positions) before attempting to fill out the implementation.

For example here, we'd have an interface for manipulating dune files: reading a file into some representation, allowing to modify that representation, and writing back to disk. For our purpose, the representation that I'd pick would be something like something like:
```
type theory = {
  theory : string;
  subdirs : int;
  direct_deps : string list;
  mutable transitive_deps : string list
}

type item =
  | Text of string
  | DepsFor of 

type t = item list * theory list
```
This would be turned back into a file by doing something like:
```
let write : Out_channel.t -> t -> unit = fun oc (items, _) ->
  let output_item item =
    match item with
    | Text(s) -> Out_channel.output_string oc s
    | DepsFor(d) ->
        let indent = String.make d.subdirs ' ' in
        List.iter (Printf.fprintf oc "\n  %s%s" indent)
          (d.direct_deps @ "; transitive deps" :: d.transitive_deps)
  in
  List.iter output_item items
```

rocq-dune-tools/lib/rewrite.ml
gmalecha-at-skylabs: I requested it. I figured that we could run it in CI the way that formatters and such get run and we would see the diff.

rocq-dune-tools/lib/workspace.ml
gmalecha-at-skylabs: I was having issues when running this via `dune exec` so this was my request. If there is a better way, that works for me.

rocq-dune-tools/lib/rewrite.ml
rlepigre-skylabs-ai: Ah, I see. We have the same issue with promoted dune files, so we should morally check for an empty git diff in CI already. The detection for whether transitive deps are up-to-date could happen via that mechanism too.

rocq-dune-tools/lib/parse.ml
gmalecha-at-skylabs: I had codex try this out and it improved things a bit, but it seems like there is enough friction between what is there and what we need that it doesn't really save much.

rocq-dune-tools/lib/rewrite.ml
gmalecha-at-skylabs: Feature removed.

pgiarrusso-sl (approved):
Approving based on design discussion, even if I haven't read the code closely.

PR #133, opened 2026-05-06
More robust implicit member generation

### LINKED ISSUE #113: Specifying POD construction/destruction is impossible
# Problem
BRiCk can't verify construction/destruction of POD types in the following examples ( https://github.com/SkyLabsAI/brick-libcpp/pull/46).

```rocq
void test_timespec_get_local() {
    std::timespec ts{};
    (void)std::timespec_get(&ts, TIME_UTC);
}

void test_timespec_dtor_bug() {
    std::timespec ts;
    (void)ts;
}
```

# Next step

I need to fill the details and hand this over to Gregory.

- It was not possible to specify destructors.
- It was not always possible to specify constructors — using `std::timespec ts{};` generated a different set of methods from `std::timespec ts;`. The constructor was only generated in one of those.

gmalecha-at-skylabs:
This is more robust than doing this in C++, but it isn't possible to get exception specifications in general because they will be inherited from bases which ultimately could use other features.

That said, these exception specifications should be computable by recursion on the type heirarchy, it just isn't something that we'd like to setup.

### DIFF
--- a/rocq-skylabs-brick/tests/implicit_members.expected
+++ b/rocq-skylabs-brick/tests/implicit_members.expected
@@ -1,0 +1,226 @@
+     = ("C1::C1()"%cpp_name,
+        Some
+          (Oconstructor
+             {|
+               c_class := "C1";
+               c_params := [];
+               c_cc := CC_C;
+               c_arity := Ar_Definite;
+               c_exception := exception_spec.Unknown;
+               c_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C1::~C1()"%cpp_name,
+        Some
+          (Odestructor
+             {|
+               d_class := "C1";
+               d_cc := CC_C;
+               d_exception := exception_spec.Unknown;
+               d_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C1::C1(const C1&)"%cpp_name,
+        Some
+          (Oconstructor
+             {|
+               c_class := "C1";
+               c_params := [("#0"%pstring, "const C1&"%cpp_type)];
+               c_cc := CC_C;
+               c_arity := Ar_Definite;
+               c_exception := exception_spec.Unknown;
+               c_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C1::C1(const C1&&)"%cpp_name, None)
+     : name * option ObjValue
+     = ("C1::operator=(const C1&)"%cpp_name,
+        Some
+          (Omethod
+             {|
+               m_return := "const C1&";
+               m_class := "C1";
+               m_this_qual := QM;
+               m_params := [("#0"%pstring, "const C1&"%cpp_type)];
+               m_cc := CC_C;
+               m_arity := Ar_Definite;
+               m_exception := exception_spec.Unknown;
+               m_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C1::operator=(C1&&)"%cpp_name,
+        Some
+          (Omethod
+             {|
+               m_return := "C1&";
+               m_class := "C1";
+               m_this_qual := QM;
+               m_params := [("#0"%pstring, "C1&&"%cpp_type)];
+               m_cc := CC_C;
+               m_arity := Ar_Definite;
+               m_exception := exception_spec.Unknown;
+               m_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C2::C2()"%cpp_name,
+        Some
+          (Oconstructor
+             {|
+               c_class := "C2";
+               c_params := [];
+               c_cc := CC_C;
+               c_arity := Ar_Definite;
+               c_exception := exception_spec.NoThrow;
+               c_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C2::~C2()"%cpp_name,
+        Some
+          (Odestructor
+             {|
+               d_class := "C2";
+               d_cc := CC_C;
+               d_exception := exception_spec.Unknown;
+               d_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C2::C2(const C2&)"%cpp_name,
+        Some
+          (Oconstructor
+             {|
+               c_class := "C2";
+               c_params := [("#0"%pstring, "const C2&"%cpp_type)];
+               c_cc := CC_C;
+               c_arity := Ar_Definite;
+               c_exception := exception_spec.Unknown;
+               c_body := Some Defaulted
+             |}))
+     : name * option ObjValue
+     = ("C2::C2(const C2&&)"%cpp_name, None)
+     : name * option ObjValue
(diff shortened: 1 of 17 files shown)


rocq-skylabs-cpp2v/src/PrintDecl.cpp
```diff
@@ -1112,10 +1119,126 @@ class PrintDecl
         return printType(Dstruct, *decl, print, cprint, ctxt);
     }
 
+    bool printImplicitMembers(const CXXRecordDecl &decl, CoqPrinter &print,
+                              ClangPrinter &cprint, const ASTContext &ctxt) {
+        (void)print;
```
pgiarrusso-sl:
```suggestion
```

rocq-skylabs-cpp2v/src/PrintDecl.cpp
```diff
@@ -1112,10 +1119,126 @@ class PrintDecl
...
+            guard::ctor _{print, "Oimplicit_move_assign"};
+            cprint.printName(print, decl);
+            printed = true;
+        }
+        if (!dtor, decl.needsImplicitDestructor()) {
+            guard::ctor _{print, "Oimplicit_dtor"};
+            cprint.printName(print, decl);
+            printed = true;
+        }
+        return printed;
```
pgiarrusso-sl, line 1200: printed isn't used, do we care or can we delete it?

rocq-skylabs-cpp2v/src/PrintDecl.cpp
```diff
@@ -1112,10 +1119,126 @@ class PrintDecl
...
+                << "::" << kind << "\n";
+        };
+
+        auto canDefine = [](const CXXMethodDecl *member) {
+            return member && member->isImplicit() && member->isDefaulted() &&
+                   !member->isDeleted() && !member->hasBody();
+        };
+
+        bool printed = false;
+        auto noteMember = [&](const CXXMethodDecl *member, bool canDeclare,
```
pgiarrusso-sl: This is used nowhere, can we delete it?

rocq-skylabs-brick/theories/lang/cpp/parser.v
```diff
@@ -237,6 +237,74 @@ Definition Dglobal_using_namespace (alias : name) : K :=
...
+                ; m_this_qual := QM
+                ; m_cc := CC_C
+                ; m_arity := Ar_Definite
+                ; m_exception := exception_spec.Unknown (* TODO: problematic? *)
+                ; m_body := Some Defaulted |} in
+  _symbols (Nscoped n $ Nop function_qualifiers.N OOEqual [arg]) ctor.
+
+Definition Oimplicit_move_assign (n : globname) : K :=
+  let is_const := false in
+  let arg := if is_const then (Tconst (Tnamed n)) else Tnamed n in
```
pgiarrusso-sl: Drop is_const?

rocq-skylabs-brick/theories/lang/cpp/parser/plugin/cpp2v_template.v
```diff
@@ -3,12 +3,16 @@
...
  *)
+Require Import Stdlib.Strings.String.
 Require Stdlib.Array.PArray.
 Require Import Stdlib.Numbers.Cyclic.Int63.PrimInt63.
 Require Import skylabs.lang.cpp.parser.
 
+Local Open Scope string_scope.
 #[local] Set Printing Universes.
 
+Definition version : string := "PLACEHOLDER".
```
pgiarrusso-sl, line 15: So if we change the binary but it produces the same .v output, you want to rebuild the .vo and all clients anyway? This seems provably unnecessary. Am I missing something?

Please drop or motivate.

rocq-skylabs-brick/tests/implicit_members.expected
```diff
@@ -0,0 +1,226 @@
...
+          (Omethod
+             {|
+               m_return := "const C4&";
+               m_class := "C4";
+               m_this_qual := QM;
+               m_params := [("#0"%pstring, "const C4&"%cpp_type)];
+               m_cc := CC_C;
+               m_arity := Ar_Definite;
+               m_exception := exception_spec.Unknown;
+               m_body := Some Defaulted
```
pgiarrusso-sl, line 222: Please ensure we have a testcase that generates a CompilerProvided body.

I didn't find one yet in this MR; I either missed it, or CompilerProvided is dead code, or test coverage is insufficient.

pgiarrusso-sl (approved):
Some comments. The test coverage is incomplete and needs to be fixed — I'm approving, but please consider that "needs changes".

rocq-skylabs-brick/theories/lang/cpp/parser/plugin/cpp2v_template.v
gmalecha-at-skylabs: The binary is an implicit dependency of the plugin. If the binary changes, they yes, everything that uses inline should be re-built.

rocq-skylabs-brick/theories/lang/cpp/syntax/decl.v
```diff
@@ -17,6 +17,7 @@ Require Import skylabs.lang.cpp.syntax.stmt.
 
 Variant OrDefault {t : Set} : Set :=
   | Defaulted
+  | CompilerProvided (_ : t)
   | UserDefined (_ : t).
```
pgiarrusso-sl, line 21: Looking at clients, maybe it would be better to have one constructor taking a boolean for both CompilerProvided and UserDefined:

`| HasBody (compiler_provided : boolean) (body : t)`.

Many clients have the same handling for both cases, except subtyping checks.

rocq-skylabs-brick/theories/lang/cpp/parser/plugin/cpp2v_template.v
pgiarrusso-sl: Hmmm we both have a point:
- for "manual" uses of `cpp.prog`, you're right — this is conservative but we don't have a better way.
- for separately generated ASTs, this is unnecessary; it seems they're affected anyway because they include the same plugin, but maybe they shouldn't.

I also _am_ seeing surprising build failures on `main` that this seems to fix, tho results are inconsistent and I don't have a full story.

rocq-skylabs-brick/theories/lang/cpp/syntax/decl.v
gmalecha-at-skylabs: My experience suggests that there is some performance improvement to fusing information like this, so I'm not going to change this at this point.

rocq-skylabs-brick/theories/lang/cpp/parser/plugin/cpp2v_template.v
gmalecha-at-skylabs: I concur that there really are two independent plugins here, and separating them would reduce dependencies, but this doesn't seem very high value IMO. Perhaps the next time that we make substantial changes to the plugin we can do this sort of change as well.

PR #134, opened 2026-05-06
[rocq-perf] Add recursive invocation detection.

rlepigre-skylabs-ai:
This should help detect issues more quickly in the future.

### DIFF
--- a/rocq-tools/bin/rocq_perf.ml
+++ b/rocq-tools/bin/rocq_perf.ml
@@ -18,6 +18,13 @@
 open Rocq_tools
 open Rocq_tools.Extra
 
+let _ =
+  try
+    ignore (Unix.getenv "IN_ROCQ_PERF");
+    panic "Error: recursive invocation of \"rocq-perf\" as %S (PATH=%s)."
+      Sys.argv.(0) (try Unix.getenv "PATH" with Not_found -> "")
+  with Not_found -> ()
+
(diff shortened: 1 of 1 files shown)


rocq-tools/bin/rocq_perf.ml
```diff
@@ -18,6 +18,13 @@
 open Rocq_tools
 open Rocq_tools.Extra
 
+let _ =
+  try
+    ignore (Unix.getenv "IN_ROCQ_PERF");
+    let path = try Unix.getenv "PATH" with Not_found -> "" in
+    panic "Error: fork bomb defused (running %S, PATH=%s)." Sys.argv.(0) path
```
pgiarrusso-sl: Nit, but I think this would be clearer:

```suggestion
    panic "Error: infinite recursion defused (running %S, PATH=%s)." Sys.argv.(0) path
```

pgiarrusso-sl (changes requested):
Just one clarification, then LGTM.

rocq-tools/bin/rocq_perf.ml
rlepigre-skylabs-ai: OK, improved the text slightly differently since defusing makes less sense without bomb.

pgiarrusso-sl (approved):

Janno (approved):

Issue #131, opened 2026-05-06
Add semantics support for 128-bit integers


gmalecha-at-skylabs:

aa755:
FWIW, codex-cli argues that the following axioms/hints are missing from BRiCk/auto:


```coq
Set Default Goal Selector "!".


Require Export skylabs.auto.cpp.tactics4.
Require Export skylabs.auto.cpp.prelude.proof.

Import linearity.

Set Warnings "+sl-impossible-patterns".


Section with_Sigma.
  Context `{Sigma : cpp_logic} {CU : genv}.

  Definition wp_init_implicit_B_local := [BWD] wp_init_implicit.

  (* Value-initializing the generated [evmc_bytes32::bytes] field materializes
     the byte array filled with zero integer bytes.  Upstream has a rule for
     [Einitlist [] (Some (Eimplicit_init ty_prim)) ...], but this AST contains
     the direct expression [Eimplicit_init (Tarray Tuchar len)] inside an
     aggregate initializer.  This narrow bridge covers that generated shape. *)
  Axiom wp_init_implicit_char_array_local :
    forall (tu : translation_unit) (rho : region) (base : ptr)
      (len : N) (Q : FreeTemps.t -> mpred),
    (base |-> type_ptrR (Tarray Tuchar len) -*
     base |-> arrayR Tuchar (primR Tuchar 1$m) (replicateN len (Vint 0)) -*
     Q FreeTemps.id)
    |-- wp_init tu rho (Tarray Tuchar len) base
          (Eimplicit_init (Tarray Tuchar len)) Q.

  Definition wp_init_implicit_char_array_local_B
      tu rho base len Q :=
    [BWD] (wp_init_implicit_char_array_local tu rho base len Q).

  (* Narrow bridge for generated zero-initialization of local
     [evmc::bytes32[N]] arrays such as [scratch] and [flat_out].  This is
     framework plumbing, not part of the MIP-8 commitment argument: the
     initialized values are overwritten before they matter semantically.
     The upstream aggregate-array constant hint expects an array constructor
     [Econstructor ctor [] ...]; here clang emits an array constructor whose
     element constructor receives an implicit [evmc_bytes32] aggregate. *)
  Axiom wp_init_bytes32_array_zero_local :
    forall (tu : translation_unit) (rho : region) (base : ptr)
      (len : N) (Q : FreeTemps.t -> mpred),
    (base |-> type_ptrR (Tarray bytes32_ty len) -*
[...]

PR #136, opened 2026-05-09
Implicit array initialization

gmalecha-at-skylabs:
As reported in https://github.com/SkyLabsAI/BRiCk/issues/131#issuecomment-4408999796.

We need to add test cases.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
@@ -18,6 +18,7 @@ Require Import skylabs.lang.cpp.logic.heap_pred.
 Require Import skylabs.lang.cpp.logic.raw.
 Require Import skylabs.lang.cpp.logic.const.
 Require Import skylabs.lang.cpp.logic.operator.
+Require Import skylabs.lang.cpp.logic.arr.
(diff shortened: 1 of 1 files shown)


rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -1660,6 +1661,13 @@ Module Type Expr.
           wp_array_init_fill ety base ls fill sz Q
       |-- wp_init (Tarray ety sz) base (Einitlist ls fill ty) Q.
 
+    Axiom wp_init_implicit_init_array : forall sz ety (base : ptr) aty default Q,
+        is_array_of aty ety ->
+        get_default ety = Some default ->
+        (base |-> type_ptrR (Tarray ety sz) -*
+         base |-> arrayR ety (primR ety 1$m) (replicateN sz default) -*
```
aa755, line 1668: is the `type_ptrR` conjunct not already entailed by the `arrayR` conjunct?

gmalecha-at-skylabs: `arrayR ety ...` implies `type_ptr ety` at each array cell, but it doesn't imply the `type_ptr (Tarray ..)`.

pgiarrusso-sl (approved):

PR #138, opened 2026-05-13
Upstreaming some proofs

gmalecha-at-skylabs:


pgiarrusso-sl (approved):

rocq-skylabs-prelude/theories/list.v
```diff
@@ -480,6 +498,18 @@ Section lists.
 End lists.
 #[global] Arguments zip_with : simpl nomatch.
 
+Lemma foldr_op_index_shift shift (la : list N) (n m : nat) f :
```
simon-skylabs: This lemma is particularly odd to me. Could we say that it's an index aware foldr homomorphism? I think that `zip` and `seq` are giving us index information. I think we could bake the index information into the assumptions about `f` and into the accumulator.

This doesn't need to block the PR but I think it would be good to leave a note to this effect.

pgiarrusso-sl: I didn't even get that far, I'm stuck at "I'd like type annotations on f and shift" like in `foldr_push_in`

gmalecha-at-skylabs: Let's make a style guide that says that we require all top-level functions to have explicit type annotations on all parameters.

simon-skylabs: I agree that type annotations are missing but there are reasonable cases where leaving out the type is more readable that putting it in. I'm not sure exactly where I'd draw the line but if you have `{A B C D}` as types and clearly used as types it's better to leave out their type.

I think it's a matter of how much mental work you need to do if you're reading the signature outside of `coqtop` in order to confidently write down the type of each argument. If the answer for a given parameter is "I gave up before finishing" that's a clear indication. I think we should recommend erring on the side of being explicit while exercising judgement.

simon-skylabs (approved):

rocq-skylabs-prelude/theories/list.v
gmalecha-at-skylabs: @pgiarrusso-sl I think that at one point you advocated for "types on definitions, no types on lemmas". Is that a good policy? Though I realize that this is a lemma and it sounds like you want types.

gmalecha-at-skylabs: I removed this upstream and will close this thread to get this merged. @pgiarrusso-sl we'll need to follow up on a separate thread.

PR #139, opened 2026-05-14
Ensure that [cpp.prog] does not have side-effects

gmalecha-at-skylabs:
Before this change, `cpp.prog ...` would `Import` files which would affect the meaning of constants, e.g. `length`.

### DIFF
--- a/rocq-skylabs-brick/tests/cpp_prog_inline.t/test.v
+++ b/rocq-skylabs-brick/tests/cpp_prog_inline.t/test.v
@@ -1,4 +1,5 @@
-Require Import skylabs.lang.cpp.parser.
+Check @length.
+
 Require Import skylabs.lang.cpp.parser.plugin.cpp2v.
 
 #[duplicates(warn)]
@@ -31,3 +32,6 @@ cpp.prog source3 prog cpp:{{
   }
 }}.
 End code.
+
+(* check that the interpretation of [length] does not change *)
+Check @length.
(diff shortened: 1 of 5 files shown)


rocq-skylabs-cpp2v/src/ToCoq.cpp
```diff
@@ -141,11 +141,11 @@ void ToCoqConsumer::toCoqModule(clang::ASTContext *ctxt,
             CoqPrinter print(fmt, /*templates*/ false, cache);
             ClangPrinter cprint(compiler_, ctxt, trace_, comment_, typedefs_);
 
-            parser(print);
             if (interactive_.has_value()) {
                 print.output() << "Section cpp_prog__" << interactive_.value()
                                << "__." << fmt::line;
             }
+            parser(print);
```
simon-skylabs, line 148: I take it the import statements put in `print` and they used to be printed before the beginning of a section which caused the side effects to persist. Could the contents of `print` be made clearer in the code?

I also see that the beginning of the section is emitted only conditionally. If `Section` is not emitted, will the side effects persist?

pgiarrusso-sl (approved):
Thank you! I think I run into this in the wild... The side effects broke a proof, but not sure I submitted a test.

rocq-skylabs-cpp2v/src/ToCoq.cpp
gmalecha-at-skylabs: The `Section` is only omitted if it is generating a file, and there the `Import` will not have effects outside the file.

rocq-skylabs-brick/tests/cpp_prog_inline.t/test.v
```diff
@@ -1,4 +1,5 @@
-Require Import skylabs.lang.cpp.parser.
+Check @length.
```
simon-skylabs, line 1: Which definition was being referred to by `length` after the import?

gmalecha-at-skylabs: PrimString.length

PR #149, opened 2026-05-19
Use [generate_project_file]

gmalecha-at-skylabs:
Status: Blocked on updating vsrocq.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,5 @@ html
 .Makefile.coq.d
 _build
 cpp2v-tests/reject/*.ERR
-_CoqProject
 _RocqProject
 .vscode/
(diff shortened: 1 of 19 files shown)


rlepigre-skylabs-ai (approved):
We should additionally be able to get rid of the `_CoqProject` generation script. But that can be done in a follow-up PR since this would require an atomic merge.

.gitignore
```diff
@@ -11,4 +11,5 @@ html
 _build
 cpp2v-tests/reject/*.ERR
 _CoqProject
+_RocqProject
```
rlepigre-skylabs-ai:
```suggestion
_RocqProject
```
Maybe we should do this, to avoid issues with both `_CoqProject` and `_RocqProject` existing.

PR #159, opened 2026-05-22
Avoid relying on (to-be-removed) deprecated module.

rlepigre-skylabs-ai:


rlepigre-skylabs-ai:
@Janno do we still need to avoid importing the Iris proof mode directly? Seems like we have a lot of places that don't respect that.

pgiarrusso-sl (approved):

PR #165, opened 2026-05-24
Hide warning about flags and paths

pgiarrusso-sl:
Right now we get a constant spurious warning for every use of `cpp.prog flags...`. The whole source code gets an annoying squiggly that interferes with reading.

We could have an opt-in flag to hide it, so users have to confirm the warning's spurious for their case. But I'm not sure it's worth it.


gmalecha-at-skylabs:
Maybe we just change the check so that it only issues a warning if it detects `-I` or `-system` or something like that.

PR #172, opened 2026-05-29
Only warn when we detect something path-like

gmalecha-at-skylabs:
This should quiet a lot of uses of flags that are harmless.

### DIFF
--- a/rocq-skylabs-brick/plugin/src/cpp2v.ml
+++ b/rocq-skylabs-brick/plugin/src/cpp2v.ml
@@ -173,38 +173,59 @@ let temp_file ?(prefix="ocaml_temp_") ?(suffix=".tmp") content =
       let _ = try Sys.remove temp_file with _ -> () in
       raise e
 
-(** Determine whether we should emit a path warning.
-   Only do this if we notice some path arguments
- *)
-let should_warn (flags : string list) : bool =
-  let exact_path_option =
-    Str.regexp "^\\(-F\\|-I\\|-idirafter\\|-iframework\\|-iframeworkwithsysroot\\|-imacros\\|-include\\|-include-pch\\|-iquote\\|-isysroot\\|-isystem\\|-ivfsoverlay\\|-resource-dir\\|--sysroot\\)$"
-  in
-  let joined_path_option =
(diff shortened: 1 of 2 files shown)


rocq-skylabs-brick/plugin/src/cpp2v.ml
```diff
@@ -173,6 +173,39 @@ let temp_file ?(prefix="ocaml_temp_") ?(suffix=".tmp") content =
...
+
+let%test "should_warn ignores empty flags" =
+  not (should_warn [])
+
+let%test "should_warn ignores flags without paths" =
+  not (should_warn ["-std=c++23";"-Wall";"-DNAME=value"])
+
+let%test "should_warn detects separated path flags" =
+  should_warn ["-I";"include"] &&
+  should_warn ["-isystem";"/usr/include"] &&
```
pgiarrusso-sl: Not required, but absolute paths shouldn't be a problem:
```suggestion
  not (should_warn ["-isystem";"/usr/include"]) &&
```

rocq-skylabs-brick/plugin/src/cpp2v.ml
```diff
@@ -173,6 +173,39 @@ let temp_file ?(prefix="ocaml_temp_") ?(suffix=".tmp") content =
...
+  should_warn ["-include";"header.hpp"] &&
+  should_warn ["--sysroot";"/opt/sysroot"]
```
pgiarrusso-sl:
```suggestion
  not (should_warn ["--sysroot";"/opt/sysroot"])
```

rocq-skylabs-brick/plugin/src/cpp2v.ml
```diff
@@ -173,6 +173,39 @@ let temp_file ?(prefix="ocaml_temp_") ?(suffix=".tmp") content =
...
+
+let%test "should_warn detects joined path flags" =
+  should_warn ["-Iinclude"] &&
+  should_warn ["-isystem/usr/include"] &&
```
pgiarrusso-sl: Same

pgiarrusso-sl (approved):

pgiarrusso-sl:
This should be squashed probably.

I changed base branch because CI against other branches doesn't rebase/merge with master, so the build was failing.

rocq-skylabs-brick/plugin/src/cpp2v.ml
gmalecha-at-skylabs: Hmmm...I guess that is correct. It just matters across systems. I can revise.

PR #178, opened 2026-06-01
Fix dune-rocqdeps bug that makes it mutate entire workspace

Janno:


gmalecha-at-skylabs (approved):
I used to use this feature and then run `make peek` at the top-level to see what repos changed. I guess now I just need to run the overall script at the top-level.

Janno:
> I used to use this feature and then run `make peek` at the top-level to see what repos changed. I guess now I just need to run the overall script at the top-level.

I am not sure what you mean. The readme states it shouldn't mutate anything outside the current directory. Did I break an actual feature?

PR #182, opened 2026-06-02
Unsupported on verbose

gmalecha-at-skylabs:
Move unsupported warnings to verbose mode (or higher)

### DIFF
--- a/rocq-skylabs-cpp2v/include/Logging.hpp
+++ b/rocq-skylabs-cpp2v/include/Logging.hpp
@@ -13,8 +13,8 @@ namespace logging {
 enum Level : int {
     FATAL = -1,
     NONE = 0,
-    UNSUPPORTED = 5,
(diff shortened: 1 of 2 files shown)


rlepigre-skylabs-ai (approved):

rlepigre-skylabs-ai (commented):
Personally, I think we should disable all non-`cpp2v`-specific warnings coming from the compiler front-end, and also make all warnings non-fatal.

I believe we do pass through the warning configuration from the CLI flags, but when ingesting a project I don't think it brings anything to have the warnings both when running the C++ compiler and when running `cpp2v`. Some warnings actually get in the way of `cpp2v`, for example the warning that triggers when processing `.hpp` files that have `#pragma once` ("should not be used in the main file"). I suspect that if the user has set `-Werror` then this warning would even be fatal.

PR #188, opened 2026-06-03
dune-rocqdeps: output actual changes, never depend on Corelib explicitly

pgiarrusso-sl:
For https://github.com/SkyLabsAI/ci/issues/172


gmalecha-at-skylabs (approved):
I'm okay with it.

PR #190, opened 2026-06-04
Introduce the C++ language version into `abi.t`

gmalecha-at-skylabs:
This isn't really part of the abi.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/syntax/abi.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/syntax/abi.v
@@ -30,6 +30,8 @@ NES.Begin abi.
     (* ^ whether or not `char` is signed or unsigned *)
   ; wchar_signed : signed
   ; byte_order : endian
+  ; lang_version : lang_version.t
+    (* ^ the C++ language version used to compile the translation unit *)
   }.
   #[only(lens,eq_dec)] derive t.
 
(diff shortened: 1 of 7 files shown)


rocq-skylabs-brick/theories/lang/cpp/syntax/abi.v
```diff
@@ -79,9 +81,9 @@ NES.Begin abi.
   (** Common Linux ABIs. Just for testing. *)
 
   Definition x64_linux : t :=
-    mkT int_rank.Ilong Signed Signed Little.
+    mkT int_rank.Ilong Signed Signed Little lang_version.Cpp14.
   Definition armle64_linux : t :=
-    mkT int_rank.Ilong Unsigned Unsigned Little.
+    mkT int_rank.Ilong Unsigned Unsigned Little lang_version.Cpp14.
```
pgiarrusso-sl: Let's default to cpp17 to keep the same semantics?

rocq-skylabs-brick/theories/lang/cpp/syntax/translation_unit.v
```diff
@@ -278,6 +278,10 @@ Record translation_unit : Type := makeTranslationUnit {
 Definition byte_order (tu : translation_unit) : endian :=
   tu.(abi).(abi.byte_order).
 
+Definition language_version (tu : translation_unit) : lang_version.t :=
+  tu.(abi).(abi.lang_version).
+Coercion language_version : translation_unit >-> lang_version.t.
```
pgiarrusso-sl: I'd recommend not using a coercion if possible? This will be very confusing.

pgiarrusso-sl (approved):

pgiarrusso-sl (approved):
This diff LGTM, and the compatibility story should be good, but please tag me if fixing CI needs significant changes.

rocq-skylabs-brick/theories/lang/cpp/syntax/translation_unit.v
gmalecha-at-skylabs: I figured that the coercion would make things a bit more future proof since we'll be able to write `evaluation_order.order_of tu xxx` rather than `evaluation_order.order_of (lang_version tu) xxx`

PR #192, opened 2026-06-04
Avoid partial output in cpp2v

gmalecha-at-skylabs:
Write the output to `foo.v.partial` and then move it to `foo.v` on success. This avoids partially constructed files that will never typecheck.

### DIFF
--- a/rocq-skylabs-cpp2v/src/ToCoq.cpp
+++ b/rocq-skylabs-cpp2v/src/ToCoq.cpp
@@ -23,7 +23,10 @@
 #include "clang/Frontend/CompilerInstance.h"
 #include "clang/Frontend/FrontendAction.h"
 #include <Formatter.hpp>
+#include <cerrno>
+#include <cstdio>
 #include <list>
+#include <system_error>
 
 // Declares clang::SyntaxOnlyAction.
 #include "SpecCollector.hpp"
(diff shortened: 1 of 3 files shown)


rlepigre-skylabs-ai (approved):
Isn't it simpler to delete the partially-generated file on failure?

I'm fine with that approach though. The only thing I wonder is whether we should try to delete the `.partial` file on failure.

pgiarrusso-sl (approved):

pgiarrusso-sl:
> Isn't it simpler to delete the partially-generated file on failure?
> 
> I'm fine with that approach though. The only thing I wonder is whether we should try to delete the `.partial` file on failure.

Good questions — the partial output might help debugging, and `rename` is atomic.

But "preserve old outputs if the build failed" might not be what we want. 

EDIT: ~~In fact, it might break `dune` caching?~~ no, I think `dune` will clear the old output first anyway.

gmalecha-at-skylabs:
For debugging, the partial output is quite helpful. I can set the PR up to delete the file by default and have it be preserved with an option (maybe reusing `-v`).

rlepigre-skylabs-ai:
It's really not a big deal if we don't delete the file, I was just wondering. Up to you!

PR #193, opened 2026-06-04
Printing for sugared functions

gmalecha-at-skylabs:
These seem to arise in C code.


pgiarrusso-sl (approved):
Approving, but could we have a testcase?

PR #198, opened 2026-06-05
Printing Unresolved Terms

gmalecha-at-skylabs:
Harden the dependent name printer

### DIFF
--- a/rocq-skylabs-cpp2v/tests/templates/dependent_name_printers.t/test.v
+++ b/rocq-skylabs-cpp2v/tests/templates/dependent_name_printers.t/test.v
@@ -1,5 +1,8 @@
 Require Import skylabs.prelude.base.
-Require Import skylabs.lang.cpp.mparser.
+Require Import skylabs.lang.cpp.syntax.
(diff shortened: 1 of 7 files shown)


rocq-skylabs-cpp2v/tests/templates/dependent_name_printers.t/test.v
```diff
@@ -0,0 +1,249 @@
+Require Import skylabs.prelude.base.
+Require Import skylabs.lang.cpp.mparser.
+Require Import test_17_cpp_templates.
+
+#[local] Open Scope pstring_scope.
```
pgiarrusso-sl: Not part of this PR, but can this be avoided with the Bind Scope commands upstream?

pgiarrusso-sl (approved):
Not sure why you do the last commit, but it doesn't affect tests here?
Everything elsd LGTM

pgiarrusso-sl:
At a second look, I can see the Rocq dependent name printer is straightforward so this makes sense.

PR #203, opened 2026-06-08
Re-organize template code

gmalecha-at-skylabs:
This is in preparation for an experiment where we include templated declarations inside non-templated declarations.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/syntax/mcore.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/syntax/mcore.v
@@ -1,14 +1,14 @@
 (*
- * Copyright (c) 2024 BlueRock Security, Inc.
+ * Copyright (c) 2024-2026 SkyLabs AI, Inc.
  * This software is distributed under the terms of the BedRock Open-Source License.
  * See the LICENSE-BedRock file in the repository root for details.
  *)
 Require Import skylabs.lang.cpp.syntax.prelude.
 Require Export skylabs.lang.cpp.syntax.core.
+Require Export skylabs.lang.cpp.syntax.decl.
 
 #[local] Set Primitive Projections.
 
-(** ** C++ with templates *)
 Notation Mname := name (only parsing).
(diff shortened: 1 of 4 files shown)


rocq-skylabs-brick/theories/lang/cpp/syntax/mcore.v
```diff
@@ -26,4 +26,111 @@ Notation Matomic_name := atomic_name (only parsing).
...
+
+(** ** Templated values *)
+Section template.
+  #[local] Set Universe Polymorphism.
+  #[local] Set Polymorphic Inductive Cumulativity.
+  #[local] Unset Auto Template Polymorphism.
+  Universe uV.
+  Context {V : Type@{uV}}.
+
+  Record template : Type@{uV} := Template {
```
pgiarrusso-sl, line 107: What's a template value?
If that comment is good, the type shouldn't be called `template`...  template_value, tvalue, or whatever shortening fits the convention. Or if those are bad names, then the comment needs clarifying

pgiarrusso-sl (approved):

rocq-skylabs-brick/theories/lang/cpp/syntax/mcore.v
```diff
@@ -26,4 +26,111 @@ Notation Matomic_name := atomic_name (only parsing).
...
+    EqDecision tinst.
+  Proof. solve_decision. Defined.
+End tinst.
+Add Printing Constructor tinst.
+#[global] Arguments tinst : clear implicits.
+#[global] Arguments TInst {_} _ _ & _ : assert.
+
+Require Import skylabs.lang.cpp.syntax.translation_unit.
+
+(** ** Templated values *)
```
gmalecha-at-skylabs:
```suggestion
(** ** Templated values
   <<int foo()>> is an [ObjValue]
   <<template<typename T> int foo(T)>> is a [template ObjValue] and would be represented as
   <<Template [Ptype "T"] (Ofunction ....(Tparam "T") ...)>>
*)
```

PR #204, opened 2026-06-09
Replace .elpi.in files with .elpi files; Remove elpidep

Janno:


Janno:
No use in prompting the model when I don't understand what I want..

gmalecha-at-skylabs:
@Janno I didn't look at this initially, but it might be worth trying to merge something like this in order to simplify things more.

PR #207, opened 2026-06-09
Fix some broken cram tests on Mac

pgiarrusso-sl:
cc @simon-skylabs

### DIFF
--- a/rocq-skylabs-cpp2v/tests/cpp2v/test_asm.t/run.t
+++ b/rocq-skylabs-cpp2v/tests/cpp2v/test_asm.t/run.t
@@ -1,4 +1,4 @@
   $ . ../../setup-cpp2v.sh
-  $ check_cpp2v test.cpp
-  cpp2v -v -check-types -o test_17_cpp.v test.cpp -- -std=c++17 2>&1 | sed 's/^ *[0-9]* | //'
+  $ CRAM_CXXFLAGS="-target x86_64-linux-gnu" check_cpp2v test.cpp
+  cpp2v -v -check-types -o test_17_cpp.v test.cpp -- -std=c++17 -target x86_64-linux-gnu 2>&1 | sed 's/^ *[0-9]* | //'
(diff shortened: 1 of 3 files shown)


rocq-skylabs-cpp2v/tests/cpp2v/test_asm.t/run.t
```diff
@@ -1,4 +1,4 @@
   $ . ../../setup-cpp2v.sh
-  $ check_cpp2v test.cpp
-  cpp2v -v -check-types -o test_17_cpp.v test.cpp -- -std=c++17 2>&1 | sed 's/^ *[0-9]* | //'
+  $ CRAM_CXXFLAGS="-target x86_64-linux-gnu" check_cpp2v test.cpp
```
simon-skylabs, line 2: In short, this test and `test_cc` only makes sense on `x86_64-linux-gnu` but the `setup-cpp2v.sh` leaves it unspecified so now you're adding the ability to specify it inside each test case, correct?

I believe before and after this change, global `(binaries ..)` declarations (e.g. in `dune-workspace`) would not affect what `cpp2v` resolves to, is this also right?

pgiarrusso-sl: > In short, this test and test_cc only makes sense on x86_64-linux-gnu but the setup-cpp2v.sh leaves it unspecified so now you're adding the ability to specify it inside each test case, correct?

Almost — they make sense _for_ x86 but it still make sense to run them on Mac.

- We _will_ still run the test on Mac.
- but we'll ask cpp2v to _cross-compile_ (or cross-verify) the program with `--target`.

> I believe before and after this change, global (binaries ..) declarations (e.g. in dune-workspace) would not affect what cpp2v resolves to, is this also right?

Probably true and desirable, but it shouldn't affect this MR, so let's discuss elsewhere.

simon-skylabs (approved):

PR #216, opened 2026-06-15
A params plugin that generates [Params] instances

gmalecha-at-skylabs:
The crucial thing about this plugin is that it cooks these definitions correctly. That is, if a parameter is added to a definition, then, during cooking, the number of parameters will be increased by the number of new parameters.


rocq-attrs/tests/test_params.v
```diff
@@ -0,0 +1,109 @@
...
+(*                                                                      *)
+(* Observable: [Check CHECK c n] succeeds iff constant [c] has a        *)
+(* registered [Params c n] instance.  The intended                      *)
+(* implementation records the arity AND registers it as a hint that is  *)
+(* #[local] inside a section (hence does not persist verbatim), while a *)
+(* corrected GLOBAL registration is installed at section close.         *)
+(************************************************************************)
+
+Require Import attrs.ParamsAttr.
+Require Import Stdlib.Classes.Morphisms.
```
pgiarrusso-sl: The user might forget to import `Params`, we should have a test for that.

In that case, we should probably pick `Stdlib.Classes.Morphisms.Params` anyway.

(The stdlib has 2 versions, in `Morphisms` and `CMorphisms`, but our codebase ignores the second, and it's identical so that seems an issue upstream.)

Janno: This should be implemented using `Register` and https://github.com/rocq-prover/rocq/blob/master/library/rocqlib.mli

gmalecha-at-skylabs:
Comments from @Janno

* Register + Rocqlib
* Filtering which commands can use the attribute, if that API allows that (need to check). (Using the attribute on Let triggers an anomaly, for example)
* Documenting the interaction with locality attributes
* Create sort/universe polymorphic instances for sort/universe polymorphic terms
* Use `Locality.Discharge` to remove the old instance and inline it

Janno (approved):

PR #217, opened 2026-06-15
Record static_assert in translation_units

### LINKED ISSUE #222: Support for static_assert

gmalecha-at-skylabs:
This provides this information to users and (I predict) will be very useful when working with architecture-independent proofs.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/parser.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/parser.v
@@ -64,7 +64,7 @@ Module Import translation_unit.
       | l :: ls' =>
           match compare x l with
           | Lt => x :: ls
-          | Eq => x :: l :: ls'
+          | Eq => ls (* remove duplicates *)
           | Gt => l :: insert x ls'
           end
       end.
--- a/rocq-skylabs-brick/theories/lang/cpp/parser.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/parser.v
@@ -51,6 +51,31 @@ Module Import translation_unit.
       end.
   End sort.
 
+  Module sort_static_assert.
+    Definition compare (x y : StaticAssert) : comparison :=
+      match PrimString.compare x.(sa_message) y.(sa_message) with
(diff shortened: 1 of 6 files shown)


rocq-skylabs-brick/theories/lang/cpp/parser.v
```diff
@@ -51,6 +51,31 @@ Module Import translation_unit.
...
+      | c => c
+      end.
+
+    Fixpoint insert (x : StaticAssert) (ls : list StaticAssert) : list StaticAssert :=
+      match ls with
+      | nil => x :: nil
+      | l :: ls' =>
+          match compare x l with
+          | Lt => x :: ls
+          | Eq => x :: l :: ls'
```
pgiarrusso-sl: Equivalently and more clearly
```suggestion
          | Lt | Eq => x :: ls
```

rocq-skylabs-brick/theories/lang/cpp/parser.v
```diff
@@ -51,6 +51,31 @@ Module Import translation_unit.
...
+          | Gt => l :: insert x ls'
+          end
+      end.
+
+    Fixpoint sort (ls : list StaticAssert) : list StaticAssert :=
```
pgiarrusso-sl, line 72: The duplication here isn't the worst but it's worth a TODO
```suggestion
    (* TODO could reuse [stdpp.sorting.merge_sort (fun x y => compare x y <> Gt)] instead *)
    Fixpoint sort (ls : list StaticAssert) : list StaticAssert :=
```

pgiarrusso-sl (approved):
LGTM but left 2 suggestions.

pgiarrusso-sl (approved):

rocq-skylabs-brick/theories/lang/cpp/parser.v
gmalecha-at-skylabs: We should be removing duplicates, so I don't think we should use `merge_sort`.

PR #221, opened 2026-06-16
Improve floating point support

Janno:
Gaps identified by the LLM that (hopefully) are addressed by this PR (and SkyLabsAI/auto#297)

- support checking rejects float casts and does not validate float operators;
- static binary-operator typing lacks usual arithmetic conversions involving floats;
- unary logical-not on floats is not supported dynamically or in automation;
- raw-byte/simple-predicate encodings for floats are missing;
- useful bit-level lemmas are missing;
- `auto` lacks ergonomic `TypeCompat`, `PrimVal`, `DefaultValue`, `BundledRep`, `to_bool`, cast-classification, and hint support for floats;
- BRiCk-level and `auto`-level float regression tests from the experiment are missing.

### DIFF
--- a/rocq-skylabs-brick/tests/float_support.v
+++ b/rocq-skylabs-brick/tests/float_support.v
@@ -91,85 +91,184 @@ Definition one_float : Expr := Efloat float_type.Ffloat f32_one.
 Definition one_double : Expr := Efloat float_type.Fdouble f64_one.
 Definition one_float128 : Expr := Efloat float_type.Ffloat128 f128_one.
 
-Example support_accepts_float_types :
-  check.type Tfloat16 = check.OK /\
-  check.type Tfloat = check.OK /\
-  check.type Tdouble = check.OK /\
+Example support_accepts_float16_type :
+  check.type Tfloat16 = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_type :
+  check.type Tfloat = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_double_type :
+  check.type Tdouble = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float128_type :
   check.type Tfloat128 = check.OK.
-Proof. repeat split; reflexivity. Qed.
+Proof. reflexivity. Qed.
 
 Example support_rejects_longdouble_type :
   check.type Tlongdouble = check.FAIL "unsupported floating width".
 Proof. reflexivity. Qed.
 
-Example support_accepts_float_literals :
-  check.expr one_float16 = check.OK /\
-  check.expr one_float = check.OK /\
-  check.expr one_double = check.OK /\
+Example support_accepts_float16_literal :
+  check.expr one_float16 = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_literal :
+  check.expr one_float = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_double_literal :
+  check.expr one_double = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float128_literal :
   check.expr one_float128 = check.OK.
-Proof. repeat split; reflexivity. Qed.
+Proof. reflexivity. Qed.
 
 Example support_rejects_longdouble_literal :
   check.expr (Efloat float_type.Flongdouble (float_value.zero float_type.Flongdouble)) =
     check.FAIL "unsupported floating width".
 Proof. reflexivity. Qed.
 
-Example support_accepts_float_casts :
-  check.expr (Ecast (Cfloat Tdouble) one_float) = check.OK /\
-  check.expr (Ecast (Cfloat Tfloat128) one_double) = check.OK /\
-  check.expr (Ecast (Cint2float Tfloat16) (Eint 0 Tint)) = check.OK /\
-  check.expr (Ecast (Cfloat2int Tint) one_float128) = check.OK /\
+Example support_accepts_float_to_double_cast :
+  check.expr (Ecast (Cfloat Tdouble) one_float) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_double_to_float128_cast :
+  check.expr (Ecast (Cfloat Tfloat128) one_double) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_int_to_float16_cast :
+  check.expr (Ecast (Cint2float Tfloat16) (Eint 0 Tint)) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float128_to_int_cast :
+  check.expr (Ecast (Cfloat2int Tint) one_float128) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_to_bool_cast :
   check.expr (Ecast Cfloat2bool one_float) = check.OK.
-Proof. repeat split; reflexivity. Qed.
+Proof. reflexivity. Qed.
 
 Definition small_enum_name : name := "Small".
 
-Example support_accepts_char_enum_float_casts :
-  check.expr (Ecast (Cint2float Tfloat) (Echar 65 Tchar)) = check.OK /\
-  check.expr (Ecast (Cfloat2int Tchar) one_float) = check.OK /\
-  check.expr (Ecast (Cint2float Tfloat) (Eenum_const small_enum_name "A")) = check.OK /\
+Example support_accepts_char_to_float_cast :
+  check.expr (Ecast (Cint2float Tfloat) (Echar 65 Tchar)) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_to_char_cast :
+  check.expr (Ecast (Cfloat2int Tchar) one_float) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_enum_to_float_cast :
+  check.expr (Ecast (Cint2float Tfloat) (Eenum_const small_enum_name "A")) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_to_enum_cast :
   check.expr (Ecast (Cfloat2int (Tenum small_enum_name)) one_float) = check.OK.
-Proof. repeat split; reflexivity. Qed.
+Proof. reflexivity. Qed.
 
-Example support_accepts_float_ops :
-  check.expr (Eunop Uplus one_float Tfloat) = check.OK /\
-  check.expr (Eunop Uminus one_float Tfloat) = check.OK /\
-  check.expr (Eunop Unot one_float Tbool) = check.OK /\
-  check.expr (Ebinop Badd one_float one_double Tdouble) = check.OK /\
-  check.expr (Ebinop Bsub one_float one_float Tfloat) = check.OK /\
-  check.expr (Ebinop Bmul one_float one_float Tfloat) = check.OK /\
-  check.expr (Ebinop Bdiv one_float one_float Tfloat) = check.OK /\
+Example support_accepts_float_unary_plus :
+  check.expr (Eunop Uplus one_float Tfloat) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_unary_minus :
+  check.expr (Eunop Uminus one_float Tfloat) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_logical_not :
+  check.expr (Eunop Unot one_float Tbool) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_add :
+  check.expr (Ebinop Badd one_float one_double Tdouble) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_sub :
+  check.expr (Ebinop Bsub one_float one_float Tfloat) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_mul :
+  check.expr (Ebinop Bmul one_float one_float Tfloat) = check.OK.
+Proof. reflexivity. Qed.
+
+Example support_accepts_float_div :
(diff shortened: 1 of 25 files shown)


rocq-skylabs-brick/tests/float_support.v
```diff
@@ -0,0 +1,296 @@
...
+  check.expr one_float128 = check.OK.
+Proof. repeat split; reflexivity. Qed.
+
+Example support_accepts_float_casts :
+  check.expr (Ecast (Cfloat Tdouble) one_float) = check.OK /\
+  check.expr (Ecast (Cfloat Tfloat128) one_double) = check.OK /\
+  check.expr (Ecast (Cint2float Tfloat16) (Eint 0 Tint)) = check.OK /\
+  check.expr (Ecast (Cfloat2int Tint) one_float128) = check.OK /\
+  check.expr (Ecast Cfloat2bool one_float) = check.OK.
+Proof. repeat split; reflexivity. Qed.
```
gmalecha-at-skylabs: In practice, I think that this testing strategy is somewhat fragile. These tests are probably okay, but for testing syntax, I much prefer writing actual C++ code using `cpp.prog ...` and then writing functions on this. This ensures that we capture the way that clang interprets syntax.

rocq-skylabs-brick/tests/float_support.v
```diff
@@ -0,0 +1,296 @@
...
+  is_true (Vfloat float_type.Ffloat f32_nan) = Some true.
+Proof. repeat split; vm_compute; reflexivity. Qed.
+
+Example float16_and_float128_bool_cases :
+  is_true (Vfloat float_type.Ffloat16 f16_neg_zero) = Some false /\
+  is_true (Vfloat float_type.Ffloat16 f16_one) = Some true /\
+  is_true (Vfloat float_type.Ffloat128 f128_neg_zero) = Some false /\
+  is_true (Vfloat float_type.Ffloat128 f128_inf) = Some true /\
+  is_true (Vfloat float_type.Ffloat128 f128_nan) = Some true.
+Proof. repeat split; vm_compute; reflexivity. Qed.
```
gmalecha-at-skylabs: Do you think it would make sense to put some of these tests closer to the actual code with `Succeed ...`? This file seems to contain a lot of tests for features that exist in different parts of the code. Sometimes this can make incremental updates difficult.

rocq-skylabs-brick/tests/float_support.v
```diff
@@ -0,0 +1,296 @@
...
+Example raw_float_intro_examples {σ : genv} :
+  raw_bytes_of_val σ Tfloat16 (Vfloat float_type.Ffloat16 f16_one)
+    (float_raw_bytes σ float_type.Ffloat16 f16_one) /\
+  raw_bytes_of_val σ Tfloat (Vfloat float_type.Ffloat f32_one)
+    (float_raw_bytes σ float_type.Ffloat f32_one) /\
+  raw_bytes_of_val σ Tdouble (Vfloat float_type.Fdouble f64_one)
+    (float_raw_bytes σ float_type.Fdouble f64_one) /\
+  raw_bytes_of_val σ Tfloat128 (Vfloat float_type.Ffloat128 f128_one)
+    (float_raw_bytes σ float_type.Ffloat128 f128_one).
+Proof. repeat split; apply raw_bytes_of_val_float_intro. Qed.
```
gmalecha-at-skylabs: Can we propose that tests such as these are split up into separate `Example`s?

rocq-skylabs-brick/theories/lang/cpp/syntax/supported.v
```diff
@@ -21,6 +23,85 @@ Section with_monad.
...
+        FAIL "unsupported pointer/floating operator"
+      else
+        match op with
+        | Badd | Bsub | Bmul | Bdiv
+        | Beq | Bneq | Blt | Ble | Bgt | Bge => OK
+        | Bmod | Band | Bor | Bxor | Bshl | Bshr | Bcmp
+        | Bdotp | Bdotip => FAIL "unsupported floating binary operator"
+        | Bunsupported _ => OK
+        end
+    else OK.
```
gmalecha-at-skylabs, line 101: This looks like it is replicating some logic. Would it be meaningful to call `convert_type_op` here? If not, could we refactor that function so that it does make sense?

rocq-skylabs-cpp2v/src/PrintExpr.cpp
```diff
@@ -963,23 +1003,43 @@ class PrintExpr : public ConstStmtVisitor<PrintExpr, void> {
...
+                break;
+            case BuiltinType::Float128:
+                if (lit->getRawSemantics() == llvm::APFloatBase::S_IEEEquad)
+                    ft = "float_type.Ffloat128";
+                break;
+            case BuiltinType::LongDouble:
+                /* The port's [Flongdouble] carrier is not the experiment's
+                   binary128 model and does not match the usual Clang x87 or
+                   double-double object encodings, so reject literals here. */
+                break;
```
gmalecha-at-skylabs, line 1029: We should be consistent about this. It seems like here we say that `long double` is not supported, but in the Rocq code we seem to say that `Flongdouble` is supported (the `supported` function is constant `true`).

I propose that we either:
1. make `long double` unsupported everywhere, or
2. introduce information into `abi.t` and then use it correctly

gmalecha-at-skylabs (changes requested):

rocq-skylabs-brick/theories/lang/cpp/syntax/preliminary.v
```diff
@@ -522,6 +567,15 @@ Module float_value.
...
     end.
 
+  Definition is_true (t : t) (v : car t) : bool :=
+    negb (is_zero v).
+
+  Lemma is_zero_zero (t : t) : is_zero (zero t) = true.
+  Proof. by destruct t. Qed.
+
+  Lemma is_true_zero (t : t) : is_true t (zero t) = false.
+  Proof. by destruct t. Qed.
```
pgiarrusso-sl:
```suggestion
  Definition is_true {t : t} (v : car t) : bool :=
    negb (is_zero v).

  Lemma is_zero_zero (t : t) : is_zero (zero t) = true.
  Proof. by destruct t. Qed.

  Lemma is_true_zero (t : t) : is_true (zero t) = false.
  Proof. by destruct t. Qed.
```

rocq-skylabs-brick/theories/lang/cpp/syntax/preliminary.v
```diff
@@ -670,6 +724,78 @@ Module float_value.
...
+
+  Lemma of_to_bits_Fdouble (f : car Fdouble) :
+    of_bits Fdouble (to_bits Fdouble f) = f.
+  Proof. apply of_to_bits. Qed.
+
+  Lemma of_to_bits_Ffloat128 (f : car Ffloat128) :
+    of_bits Ffloat128 (to_bits Ffloat128 f) = f.
+  Proof. apply of_to_bits. Qed.
+
+  Lemma to_bits_inj ft : Inj (=) (=) (to_bits ft).
```
pgiarrusso-sl:
```suggestion
  #[global] Instance to_bits_inj ft : Inj (=) (=) (to_bits ft).
```

rocq-skylabs-brick/theories/lang/cpp/syntax/preliminary.v
```diff
@@ -670,6 +724,78 @@ Module float_value.
...
+  Proof. apply of_to_bits. Qed.
+
+  Lemma of_to_bits_Fdouble (f : car Fdouble) :
+    of_bits Fdouble (to_bits Fdouble f) = f.
+  Proof. apply of_to_bits. Qed.
+
+  Lemma of_to_bits_Ffloat128 (f : car Ffloat128) :
+    of_bits Ffloat128 (to_bits Ffloat128 f) = f.
+  Proof. apply of_to_bits. Qed.
+
```
pgiarrusso-sl: Seem unnecessary?
```suggestion
```

rocq-skylabs-brick/theories/lang/cpp/syntax/preliminary.v
```diff
@@ -670,6 +724,78 @@ Module float_value.
...
+  Proof. intros Hz. apply to_of_bits. exact Hz. Qed.
+
+  Lemma to_of_bits_Fdouble z :
+    (0 <= z < 2 ^ 64)%Z -> to_bits Fdouble (of_bits Fdouble z) = z.
+  Proof. intros Hz. apply to_of_bits. exact Hz. Qed.
+
+  Lemma to_of_bits_Ffloat128 z :
+    (0 <= z < 2 ^ 128)%Z -> to_bits Ffloat128 (of_bits Ffloat128 z) = z.
+  Proof. intros Hz. apply to_of_bits. exact Hz. Qed.
+
```
pgiarrusso-sl, line 768: Questionable?

rocq-skylabs-brick/theories/lang/cpp/syntax/preliminary.v
```diff
@@ -465,6 +487,29 @@ Module float_type.
...
+  Lemma bit_width_Ffloat16 : bit_width Ffloat16 = 16%Z.
+  Proof. reflexivity. Qed.
+  Lemma bit_width_Ffloat : bit_width Ffloat = 32%Z.
+  Proof. reflexivity. Qed.
+  Lemma bit_width_Fdouble : bit_width Fdouble = 64%Z.
+  Proof. reflexivity. Qed.
+  Lemma bit_width_Flongdouble : bit_width Flongdouble = 79%Z.
+  Proof. reflexivity. Qed.
+  Lemma bit_width_Ffloat128 : bit_width Ffloat128 = 128%Z.
+  Proof. reflexivity. Qed.
```
pgiarrusso-sl: Useful? These might be TBH, but then I might write `bit_width` as a match so `simpl` does the job?

```suggestion
```

pgiarrusso-sl (commented):
Some comments. Specs and proofs seem a bit dodgy.

rocq-skylabs-brick/theories/lang/cpp/syntax/preliminary.v
Janno: In what sense? Is the question if these are useful?

rocq-skylabs-cpp2v/src/PrintExpr.cpp
Janno: > 1. make long double unsupported everywhere

Seems like the right choice for this PR here.

rocq-skylabs-brick/theories/lang/cpp/syntax/supported.v
Janno: `convert_type_op` needs a TU. The LLM is proposing to use `convert_type_float_op` instead to replace the inner `match`. Would that help?
```
  Definition check_float_binop (op : BinOp) : M :=
    match convert_type_float_op op Tfloat with
    | Some _ => OK
    | None => FAIL "unsupported floating binary operator"
    end.

  Definition check_binop (op : BinOp) (lhsT rhsT resT : type) : M :=
    (match op with
     | Bunsupported msg => FAIL msg
     | _ => OK
     end) <+>
    if is_float_type lhsT || is_float_type rhsT || is_float_type resT then
      check_float_type_if_present lhsT <+>
      check_float_type_if_present rhsT <+>
      check_float_type_if_present resT <+>
      if is_pointer lhsT || is_pointer rhsT then
        FAIL "unsupported pointer/floating operator"
      else
        match op with
        | Bunsupported _ => OK
        | _ => check_float_binop op
        end
    else OK.
```

rocq-skylabs-cpp2v/src/PrintExpr.cpp
Janno: Hm, I realize that I don't really know what this means. I asked the agent to mark unsupported items unsupported consistently across the entire stack and it changed supported.v to check for the exact float type (which main apparently does not) and that now breaks some tests that rely on `long double` being considered supported. Is that the right way to go? Do I just fix the tests?

gmalecha-at-skylabs: If we're deciding to not support `long double`, then updating tests that state that `long double` is supported seems like the right thing to do.

rocq-skylabs-brick/theories/lang/cpp/syntax/supported.v
gmalecha-at-skylabs: A better solution is probably to thread the actual TU through the `supported` check and then pass it here. Then this becomes "supported in a context".

rocq-skylabs-brick/theories/lang/cpp/logic/raw.v
```diff
@@ -65,6 +65,49 @@ Axiom raw_byte_of_int_eq : ∀ {σ : genv} sz x rs,
...
+(** Reverse raw-byte reinterpretation is axiomatized at this abstraction level:
+    [raw_byte] deliberately does not expose an injective byte-to-[N] view, so
+    the existing integer raw-byte characterization is not enough to reconstruct
+    the byte list needed by [float_raw_bytes].  The compatibility premise
+    excludes the current [Flongdouble] model, whose Flocq bit encoding does not
+    occupy all 16 object bytes. *)
+Axiom raw_bytes_of_val_unsigned_bits_to_float : forall {σ : genv} sz ft z rs,
+  float_bits_compatible sz ft ->
+  raw_bytes_of_val σ (Tnum sz Unsigned) (Vint z) rs ->
+  raw_bytes_of_val σ (Tfloat_ ft) (Vfloat ft (float_value.of_bits ft z)) rs.
```
gmalecha-at-skylabs, line 109: This axiom seems quite awkward, and I'm wondering what it is really achieving.

rocq-skylabs-brick/theories/lang/cpp/syntax/supported.v
gmalecha-at-skylabs: After reviewing it seems like this is putting the logic in the wrong place. I have clarified this in the comments of the file.

gmalecha-at-skylabs (approved):

PR #228, opened 2026-06-18
Fix specialize_products bug with evars

Janno:


Janno:
It's slower but also more correct. I cannot figure out why it would be any slower in files that don't use evars.

rlepigre-skylabs-ai (approved):

PR #233, opened 2026-06-26
Upstream list lemmas

simon-skylabs:


rocq-skylabs-prelude/theories/base.v
```diff
@@ -54,6 +54,19 @@ Lemma iff_and (P Q R : Prop) :
   (P ∧ Q) <-> (P ∧ R).
 Proof. tauto. Qed.
 
+(** This is meant for proofs by WLOG. When proving <<P <-> Q>>, one can assume <<X>> if both sides
```
pgiarrusso-sl, line 57: is that with the `wlog` tactic or something else?

rocq-skylabs-prelude/theories/list_numbers.v
```diff
@@ -1222,11 +1342,21 @@ Definition replicateZ {A} (n : Z) := (replicateN (A :=A) (Z.to_N n)).
...
+
 
   Lemma lengthZ_eq_sub_iff {A} (m n : Z) (xs : list A) :
     lengthZ xs = m - n <-> lengthN xs = Z.to_N (m - n) ∧ n ≤ m.
   Proof. lia. Qed.
 
+  Lemma lengthN_eq_iff_lengthZ {A} (xs : list A) (n : N) : lengthN xs = n <-> lengthZ xs = n.
+  Proof. by lia. Qed.
+
+  #[local] Hint Rewrite @lengthN_eq_iff_lengthZ : lift_NtoZ.
```
pgiarrusso-sl, line 1365: is `#[local]` intended?

rocq-skylabs-prelude/theories/list_numbers.v
```diff
@@ -2268,3 +2779,69 @@ Section replicateZ_lemmas.
...
+      have {}Hi : i `max` 0 = i by lia.
+      move: Hi Hxs => + /split_atZ_iff_lookupZ => -> [Hlen ->].
+      rewrite lookupZ_app bool_decide_eq_false_2 //.
+      by rewrite Hlen Z.sub_diag lookupZ_cons bool_decide_eq_true_2.
+  Qed.
+
+End split_atZ_lemmas.
+
+Definition Forall_simpl {A} :=
+  (Forall_nil (A := A),Forall_singleton (A := A), Forall_cons (A := A), Forall_app (A := A)).
```
pgiarrusso-sl:
```suggestion
  (Forall_nil (A := A), Forall_singleton (A := A), Forall_cons (A := A), Forall_app (A := A)).
```

pgiarrusso-sl (approved):

rocq-skylabs-prelude/theories/tactics/num_lifting.v
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * This software is distributed under the terms of the BedRock Open-Source License.
+ * See the LICENSE-BedRock file in the repository root for details.
+ *)
+
+(** * This file defines the tactic [lift_NtoZ]
+
+    [lift_NtoZ] is used in combination with a database of rewriting rules to lift lemmas about type
+    [N] -- in particular when they are used by list functions as is done in
+    [skylabs.prelude.list_numbers] -- and rephrase them in terms of [Z].
```
pgiarrusso-sl: I propose to add a TODO and defer clarifications.

```suggestion
    [lift_NtoZ] is used in combination with a database of rewriting rules to lift lemmas from type
    [N] to type [Z] -- in particular when they are used by list functions as is done in
    [skylabs.prelude.list_numbers].
    
    TODO: extend documentation.
```

rocq-skylabs-prelude/theories/base.v
simon-skylabs: It is with the `wlog` tactic. Consider for example a proof state like this:

```
x0 x1 : A
xs : list A
n : Z
----------
xs !! n = Some x0 /\ P <-> xs !! n = Some x1 /\ Q
```

Certain lemmas that you may be trying to use could be relying on more assumptions than are available, let's say you need `0 ≤ n < lengthZ xs` for instance. Informally, you'd be arguing "without loss of generality, we can assume `0 ≤ n < lengthZ xs`":

```
wlog : / 0 ≤ n < lengthZ xs.
{ apply iff_of_both_implies.
   - (* |-- xs !! n = Some x0 /\ P -> 0 ≤ n < lengthZ xs *)
      ...
   - (* |-- xs !! n = Some x1 /\ Q -> 0 ≤ n < lengthZ xs *)
      ... }
move => Hbounds.
(* 
...
Hbounds : 0 ≤ n < lengthZ xs
----------
xs !! n = Some x0 /\ P <-> xs !! n = Some x1 /\ Q *)
```

In `list_numbers.v`, `lookupZ_takeZ_Some` and `lookupZ_dropZ_Some` are examples using this pattern.

rocq-skylabs-prelude/theories/list_numbers.v
simon-skylabs: It is needed because we're in a section. I think you may be asking whether we can export those hints more globally and I think that would be a good idea. I'll add that at the end of the section.

rocq-skylabs-prelude/theories/tactics/num_lifting.v
```diff
@@ -0,0 +1,113 @@
+(*
+ * Copyright (C) 2022 BlueRock Security, Inc.
```
gmalecha-at-skylabs:
```suggestion
 * Copyright (C) 2026 SkyLabs AI, Inc.
```

rocq-skylabs-prelude/theories/list_numbers.v
```diff
@@ -49,6 +52,20 @@ Definition rotateN {A} n xs :=
...
+  | x :: xs =>
+      match n with
+      | N0 =>
+          ([], x :: xs)
+      | Npos p =>
+        let (ys0, ys1) := split_atN (Pos.pred_N p) xs in
+        (x :: ys0, ys1)
+      end
+  end .
+#[global] Hint Opaque split_atN : typeclass_instances sl_opacity.
```
gmalecha-at-skylabs, line 70: Does the `sl_opacity` database exist here? I thought that was introduced downstream.

rocq-skylabs-prelude/theories/list_numbers.v
```diff
@@ -734,6 +763,36 @@ Section listN.
...
+      by rewrite lengthN_takeN N.min_l // takeN_dropN.
+    - move => [Hlen_ys Hxs_ys_app].
+      have Hxs_ys_app0 := f_equal (takeN n) Hxs_ys_app.
+      have Hxs_ys_app1 := f_equal (dropN n) Hxs_ys_app.
+      move: Hxs_ys_app0 Hxs_ys_app1; rewrite takeN_app dropN_app.
+      have ? : lengthN ys0 ≤ n by lia.
+      move: Hlen_ys => ->; rewrite N.sub_diag.
+      rewrite takeN_zero (takeN_lengthN _ ys0) // right_id.
+      rewrite dropN_zero (dropN_lengthN _ ys0) // left_id.
+  Qed.
```
gmalecha-at-skylabs, line 804: These should probably move closer to `split_atN`.

rocq-skylabs-prelude/theories/list_numbers.v
```diff
@@ -49,6 +52,20 @@ Definition rotateN {A} n xs :=
...
+  | [] => ([], [])
+  | x :: xs =>
+      match n with
+      | N0 =>
+          ([], x :: xs)
+      | Npos p =>
+        let (ys0, ys1) := split_atN (Pos.pred_N p) xs in
+        (x :: ys0, ys1)
+      end
+  end .
```
gmalecha-at-skylabs: Is the idea that we consider this a more efficient implementation of `(firstn ..., skipn ...)`? If so, perhaps we should make it tail recursive as well?

I don't think this is a blocker, but I'm kind of tempted to drop this function (from the upstream) unless we think it is beneficial for performance. Otherwise it just adds another symbol that clients need to think about when to use.

gmalecha-at-skylabs (approved):
I think some changes would be good, but I don't need to review again.

rocq-skylabs-prelude/theories/list_numbers.v
pgiarrusso-sl: we have it in BRiCk.

rocq-skylabs-prelude/theories/list_numbers.v
simon-skylabs: Since this is an upstream to unblock moving stuff to the simp package, I'd like minimize the downstream changes involved but I can make it tail recursive.

rocq-skylabs-prelude/theories/list_numbers.v
simon-skylabs: The style in this file puts definitions next to each other and groups related lemmas. I think the file could stand to be broken up into sections but I avoided doing to much reorganizing. I could add the sections before merging if you prefer.

rocq-skylabs-prelude/theories/list_numbers.v
simon-skylabs: ~~No I think I need to move that somewhere in `auto`.~~

Ah! Look at that! `skylabs.prelude.tactics.base_dbs.`

PR #240, opened 2026-06-30
Bug fixes for dependent name printer

### LINKED ISSUE #234: Assertion failed: print.templates(), function operator()
While trying to process a new C++ project, I encountered 16 files on which I got the following error:
```
Assertion failed: print.templates(), function operator(), file $WORKSPACE/_build/default/fmdeps/BRiCk/rocq-skylabs-cpp2v/src/ClangPrinter.cpp, line 154.
```

gmalecha-at-skylabs:
This also features additional tests and fixes that arose due to the goal.

Closes #234

### DIFF
--- a/rocq-skylabs-brick/AGENTS.md
+++ b/rocq-skylabs-brick/AGENTS.md
@@ -1,0 +1,6 @@
+# AGENTS.md
+
+- In Rocq tests that mention C++ names or types, prefer the string notations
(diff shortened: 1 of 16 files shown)


pgiarrusso-sl (approved):
The tests look reasonable; I only skimmed the actual implementation but we should get it in anyway.

PR #249, opened 2026-07-02
Split cpp.prog into separate plugins

### LINKED ISSUE #239: Split cpp.prog into separate plugins
# Context
We really have 2 commands, combined in `cpp.prog` — one used just by `cpp2v` output ASTs (call it `cpp.ast`), and one used by users.

This is 99% fine but with separate commands and plugins:
- the AST plugin wouldn't depend on the cpp2v binary hash and on having clang installed — so it could be built again once for all clang versions.
- syntax errors might improve — `cpp.prog source` gives the unhelpful

```
Error:
Syntax error: 'defns' or 'abi' or [cpp_flags] expected after [ident] (in [command]).
```
where `defns` and `abi` belong to `cpp.ast`.

# Proposal

The two commands should live in separate plugins and opam packages.
- The AST plugin should live in `rocq-skylabs-brick` like today; that should not depend on `rocq-skylabs-cpp2v` any more. This should be built once and for all for all versions.
- The C++ program plugin should depend on that _and_ on `rocq-skylabs-cpp2v`.
- The `cpp2v` generator should switch to `cpp.ast`.
- Identify and review any tests in `rocq-skylabs-brick` that need `cpp.prog`. They could move with the plugin or be rephrased without — to be determined.

pgiarrusso-sl:
Fix https://github.com/SkyLabsAI/BRiCk/issues/239.

Atomic with:

https://github.com/SkyLabsAI/auto/pull/338
https://github.com/SkyLabsAI/auto-docs/pull/22
https://github.com/SkyLabsAI/skylabs-fm/pull/286

https://github.com/SkyLabsAI/psi-verifier/pull/1344
https://github.com/SkyLabsAI/psi_PROVER.data/pull/160


rocq-skylabs-brick-cpp-prog/tests/cpp_prog_inline.v
```diff
@@ -1,6 +1,6 @@
 Check @length.
 
-Require Import skylabs.lang.cpp.parser.plugin.cpp2v.
+Require Import skylabs.prog.cpp.plugin.
```
gmalecha-at-skylabs, line 3: Nit. I propose a different name for this import. Is it possible to bundle the visible import as `skylabs.lang.cpp.prog`?

gmalecha-at-skylabs (approved):

rocq-skylabs-brick-cpp-prog/tests/cpp_prog_inline.v
pgiarrusso-sl: If it's in skylabs.lang.cpp, it must be the same dune theory.

gmalecha-at-skylabs: It seems awkward to get almost everything from one path and then 1 thing from a very different path. That seems to add complexity.

But if you think the benefits outweigh the costs, it is fine.

gmalecha-at-skylabs:
I'm quite surprised there seems to be such a consistent performance improvement. Any idea what is the cause of that? Is it just that the plugin is a little bit smaller?

pgiarrusso-sl:
> I'm quite surprised there seems to be such a consistent performance improvement. Any idea what is the cause of that? Is it just that the plugin is a little bit smaller?

There are too many changes here to be sure.
- For instance, downstream changes show that we print fewer backtraces. That's because we dropped `inline_tests`, which enables backtraces. The automation plugin uses `inline_tests`, but IIRC we changed it to force-disable backtraces. Here we didn't.

rocq-lens/Lens/Elpi/dune
```diff
@@ -3,11 +3,9 @@
  (generate_project_file)
```
pgiarrusso-sl, line 3: TODO (not here)
~~Before merge, update the `dune` generators in NOVA and bhv to list the new dependency where needed.~~ they don't need cpp.prog.

rocq-skylabs-brick-cpp-prog/tests/cpp_prog_inline.v
rlepigre-skylabs-ai: I agree it seems awkward, but let's fix this later (this should be much easier when we start using the new Rocq package format).

rlepigre-skylabs-ai:
Closing in favour of https://github.com/SkyLabsAI/BRiCk/pull/326 and friends.

PR #256, opened 2026-07-08
Add complexity estimator library with Ltac2 bindings

Janno:
I will squash all the commits before this gets in. The base commit is the state from a while ago and all the extra commits are polishing that happened yesterday and today.


guesstimato/.pi/tasks/tasks-019ed9ca-3d50-7a45-8d8a-dd2f1e5ddf6e.json
```diff
@@ -0,0 +1,41 @@
+{
```
rlepigre-skylabs-ai: Is this file meant to be committed?

guesstimato/README.md
```diff
@@ -0,0 +1,108 @@
...
+dune exec guesstimato -- fit timings.csv
+# disable default problem-size normalization
+# dune exec guesstimato -- fit --normalize-samples=false timings.csv
+# inspect parameters in the normalized fitted coordinate
+# dune exec guesstimato -- fit --fit-parameter-scale=normalized timings.csv
+# include comparisons and their rationale notes
+# dune exec guesstimato -- fit --print-comparisons=all --verbose timings.csv
+# check that the best fit is polynomial-2; exits 0 on match, 1 on mismatch
+# dune exec guesstimato -- assert polynomial-2 timings.csv
+```
```
rlepigre-skylabs-ai: It might be better to focus on how the tool is called, rather than the details of how to build it / run it in dune. If you think it's useful to say it, I would just make it a quick comment at the top that you can run the tool with `dune exec -- guesstimato [args]` if you don't want to install it, but then only show commands without the `dune exec --` prefix.

I also don't quite understand why most commands are commented out here,

guesstimato/README.md
```diff
@@ -0,0 +1,108 @@
...
+and the number of distinct problem sizes minus one, and requires at least three
+observations overall.
+
+If dune is installed only inside opam, use:
+
+```sh
+opam exec -- dune build
+opam exec -- dune exec guesstimato -- fit timings.csv
+opam exec -- dune exec guesstimato -- assert polynomial-2 timings.csv
+```
```
rlepigre-skylabs-ai: I'd just remove this, I don't think it brings much.

guesstimator/lib/core.mli
```diff
@@ -0,0 +1,201 @@
...
+    identifiable polynomial or quasi-polynomial degrees in [estimate.fits].
+    Requested degrees that cannot be fitted are ignored. This is useful for
+    callers that need to compare a specific degree against the selected model
+    even when the normal degree search would prune it. *)
+val estimate :
+  ?alpha:float ->
+  ?include_polynomial_degrees:int list ->
+  ?include_quasi_polynomial_degrees:int list ->
+  sample list ->
+  (estimate, string) result
```
rlepigre-skylabs-ai, line 275: It'd be good to use documentation comments consistently, and to add some blank line between declarations.

guesstimator/lib/guesstimato.ml
```diff
@@ -0,0 +1,2 @@
+include Core
+module Api = Api
```
rlepigre-skylabs-ai, line 2: This shouldn't be useful since the library is already wrapped. Or am I missing something?

guesstimato/ltac2-guesstimato/plugin/dune
```diff
@@ -0,0 +1,4 @@
+(library
+ (name ltac2_guesstimato_plugin)
+ (public_name ltac2-guesstimato.plugin)
+ (libraries rocq-runtime.plugins.ltac2 guesstimato))
```
rlepigre-skylabs-ai: You need to add an empty `.mlpack` file as usual, so that dune knows that this is a Rocq plugin.

ltac2-guesstimator/theories/Guesstimator.v
```diff
@@ -0,0 +1,256 @@
+Require Import Ltac2.Ltac2.
+
+Ltac2 Type api_error := [..].
```
rlepigre-skylabs-ai, line 13: I don't quite get why everything is so extensible in the API.

ltac2-guesstimator/theories/Guesstimator.v
```diff
@@ -0,0 +1,256 @@
...
+
+    - [holdout] enables k-fold round-robin hold-out validation.
+    - [holdout_tail] enables tail hold-out validation on the largest problem
+      sizes.
+    - [holdout_folds] is the number of round-robin folds used when [holdout] is
+      [true].
+    - [holdout_tail_fraction] is the fraction of largest problem sizes held out
+      when [holdout_tail] is [true].
+    - [holdout_stability_threshold] is the required fraction of hold-out runs
+      that must select the reference model before the fit is considered stable. *)
```
rlepigre-skylabs-ai, line 273: Given the number of parameters, I wonder if this should be a record with a default value that we can tweak using `{default with ...}` syntax.

guesstimato/ltac2-guesstimato/theories/Guesstimato.v
```diff
@@ -0,0 +1,256 @@
...
+    unstable, or [ApiError error] when fitting itself fails. Use
+    [api_error_to_string] for the currently known string error case. *)
+Ltac2 @ external run_assert :
+  holdout_options -> bool -> float -> complexity -> float sample list -> assert_result api_result :=
+  "ltac2_guesstimato" "run_assert".
+
+(** [fit_result_normalized result] returns the [normalize_samples] flag used to
+    produce [result]. *)
+Ltac2 @ external fit_result_normalized : fit_result -> bool :=
+  "ltac2_guesstimato" "fit_result_normalized".
```
rlepigre-skylabs-ai: I find it pretty hard to navigate the API, given the number of functions and associated arguments. I wonder if the main operations should not be separated to a different Rocq doc section.  At a quick glance, I can't understand what function I should be looking at first.

Also, the documentation seems to assume that the reader knows about quite a few concepts, so that's another thing that's not helping me follow along.

rlepigre-skylabs-ai (approved):
This looks good to me, but that's a lot of code to look at so I only skimmed it. I think the APIs could use some polishing, as they look pretty complicated to me. Not sure if it's an artifact of LLM-generation, or if there is a good reason to have so many different types and options. I can't tell which it is without spending significantly more time on the review.

guesstimato/dune-project
```diff
@@ -0,0 +1,27 @@
+(lang dune 3.22)
+(using rocq 0.12)
+(cram enable)
+(name guesstimato)
```
rlepigre-skylabs-ai: I forgot to ask, why `guesstimato`, and not `guesstimator`?

Janno: Good question. I like your version better so I'll rename it.

guesstimator/lib/guesstimato.ml
Janno: This was the main part that I was not sure about. I think the LLM was going for one entry point in the form of a module with the same name as the library. Is that a standard OCaml convention?

ltac2-guesstimator/theories/Guesstimator.v
Janno: This was an attempt to see how easy it would be to make everything forwards compatible with new error types, new options, etc. It's easy enough to undo if you think it is too much. I thought it was nice to be defensive about these things.

guesstimator/lib/guesstimato.ml
rlepigre-skylabs-ai: The OCaml convention, at least under dune, seems to be only to wrap the library. This means that you get `Guestimato.Core` and `Guestimato.Api`, which you can decide to open individually.

With what you have, you'd do `open Guestimato.Guestimato`, which would give you access to whatever is in `Guestimato.Core`, and also give you an `Api` module.

With what I suggest, you'd need to do `open Guestimato.Core` to get access to `Core`, and do `module Api = Guestimato.Api` if you wanted an `Api` module.

I believe I prefer the latter approach, but up to you. I just find weird the `Guestimato.Guestimato`.

ltac2-guesstimator/theories/Guesstimator.v
rlepigre-skylabs-ai: I see. I guess it's good to future-proof, but the downside is that the API gets more complicated.

I'll leave it up to you if you thing it's likely that the API will evolve in a compatible way in the future.

ltac2-guesstimator/theories/Guesstimator.v
Janno: I bundled these but I am not using records (for reasons of forwards compatibility). Probably overkill but I am trying to be consistent about it.

PR #258, opened 2026-07-08
Extend typechecking for Econstructor/Eglobal

pgiarrusso-sl:
The purpose of this PR is to make the type checker more complete.

### DIFF
--- a/rocq-skylabs-brick/tests/typed_decltype_checks.v
+++ b/rocq-skylabs-brick/tests/typed_decltype_checks.v
@@ -1,22 +1,6 @@
 Require Import skylabs.lang.cpp.syntax.typed.
 Require Import skylabs.lang.cpp.cpp.
-Require Import skylabs.lang.cpp.parser.plugin.cpp2v.
-
-#[duplicates(error)]
-cpp.prog source flags "-std=c++20" prog cpp:{{
-  struct C {
-    int f;
-    C(int);
-  };
-
-  struct D {
-    D();
-  };
-
-  void not_ctor(C*, int) {}
-
-  int global_int;
-}}.
+Require Import skylabs.lang.cpp.parser.
 
 Definition C_name : name := Nglobal (Nid "C").
 Definition D_name : name := Nglobal (Nid "D").
@@ -27,6 +11,37 @@ Definition global_int_name : name := Nglobal (Nid "global_int").
 Definition C_field_f : name := Nscoped C_name (Nid "f").
 Definition C_field_missing : name := Nscoped C_name (Nid "missing").
 Definition missing_global : name := Nglobal (Nid "missing").
+
+Definition C_struct : Struct :=
+  Build_Struct nil
+    (mkMember (Nid "f") Tint false None {| li_offset := 0 |} :: nil)
+    nil nil (Nscoped C_name Ndtor) true None POD 4 4.
+
+Definition D_struct : Struct :=
+  Build_Struct nil nil nil nil (Nscoped D_name Ndtor) true None POD 1 1.
+
+Definition C_ctor : Ctor :=
+  Build_Ctor C_name (("x"%pstring, Tint) :: nil) CC_C Ar_Definite
+    exception_spec.Unknown None.
+
+Definition D_ctor : Ctor :=
+  Build_Ctor D_name nil CC_C Ar_Definite exception_spec.Unknown None.
+
+Definition not_ctor : Func :=
+  Build_Func Tvoid (("self"%pstring, Tptr (Tnamed C_name)) :: ("x"%pstring, Tint) :: nil)
+    CC_C Ar_Definite exception_spec.Unknown None.
+
+Definition source : translation_unit :=
+  Eval vm_compute in
+    fst (parser.translation_unit.list_decls
+      (parser.Dstruct C_name (Some C_struct) ::
+       parser.Dstruct D_name (Some D_struct) ::
+       parser.Dconstructor C_ctor_int C_ctor ::
+       parser.Dconstructor (Nscoped D_name (Nctor nil)) D_ctor ::
+       parser.Dfunction not_ctor_name not_ctor ::
+       parser.Dvariable global_int_name Tint global_init.NoInit ::
+       nil)
+      abi.abi_default).
 
 Goal exists ov, source.(symbols) !! C_ctor_int = Some ov.
 Proof. vm_compute. eauto. Qed.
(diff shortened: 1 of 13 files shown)


rocq-skylabs-brick/tests/typed_decltype_checks.v
```diff
@@ -0,0 +1,86 @@
...
+
+Definition C_name : name := Nglobal (Nid "C").
+Definition D_name : name := Nglobal (Nid "D").
+Definition C_ctor_int : name := Nscoped C_name (Nctor (Tint :: nil)).
+Definition not_ctor_name : name :=
+  Nglobal (Nfunction function_qualifiers.N "not_ctor" (Tptr (Tnamed C_name) :: Tint :: nil)).
+Definition global_int_name : name := Nglobal (Nid "global_int").
+Definition C_field_f : name := Nscoped C_name (Nid "f").
+Definition C_field_missing : name := Nscoped C_name (Nid "missing").
+Definition missing_global : name := Nglobal (Nid "missing").
```
gmalecha-at-skylabs, line 13: These should all be written with notation.

rocq-skylabs-brick/tests/typed_decltype_checks.v
```diff
@@ -0,0 +1,86 @@
...
+Goal exists ov, source.(symbols) !! not_ctor_name = Some ov.
+Proof. vm_compute. eauto. Qed.
+
+Goal trace.runO (typed.decltype.check_tu source) = Some tt.
+Proof. vm_compute. reflexivity. Qed.
+
+Goal trace.runO (typed.decltype.of_expr source
+       (Econstructor C_ctor_int (Eint 0 Tint :: nil) (Tnamed C_name))) =
+     Some (Tnamed C_name).
+Proof. vm_compute. reflexivity. Qed.
```
gmalecha-at-skylabs, line 58: IMO, it is a bit nicer to generate the AST from Clang than to write it explicitly.

rocq-skylabs-brick/theories/lang/cpp/syntax/typed.v
```diff
@@ -83,7 +83,8 @@ Module decltype.
   Section with_lang.
     Record ext_tu : Set :=
     { ext_symbols : name -> option decltype
-    ; ext_types : name -> option GlobDecl }.
+    ; ext_types : name -> option GlobDecl
+    ; ext_values : name -> option ObjValue }.
```
gmalecha-at-skylabs: I think that this is unnecessary. If we wanted this, then we can basically replace `ext_tu` with `translation_unit`.

rocq-skylabs-brick/theories/lang/cpp/syntax/typed.v
```diff
@@ -556,14 +614,9 @@ Module decltype.
...
             mret $ Tenum n
         | Eglobal nm ty =>
-            (* TODO NAMES TYPING:
-               1. fields are not included, these are necessary for <<sizeof()>> (see cxx/bitvector)
-             *)
-            (*
-            let* from_env := type_of_global nm in
-            trace ("from_env = ", from_env))%bs $
+            let* from_env := type_of_global_or_field nm in
+            trace ("from_env = "%bs, from_env) $
```
gmalecha-at-skylabs, line 625: I believe that the old comment was simply out of date. `Eglobal_member` was probably introduced for this purpose.

rocq-skylabs-brick/theories/lang/cpp/syntax/typed.v
```diff
@@ -743,8 +796,10 @@ Module decltype.
             const Tsize_t <$> guard (t = Tsize_t)
         | Eoffsetof _ _ t => mret t
         | Econstructor f es t =>
-            (* TODO: check the type of the constructor [f] *)
+            let* ctor := constructor_decl f in
             let* tes := traverse (T:=eta list) of_expr es in
+            let* _ := check_args ctor.(c_arity) (snd <$> ctor.(c_params)) tes in
```
gmalecha-at-skylabs: The issue showing up in CI seems to be related to this line. We are initializing an `int` from an `int&`. I have not found a way to reproduce this in explicit code, but it does show up in compiler written code. If that is the case, then we have two choices:

1. Fix the output cpp2v by inserting the appropriate implicit cast in the AST explicitly.
2. Fold this into the semantics and extend initialization to support introducing implicit reads.

the first seems a little bit easier and more uniform, so I might opt for that, but there may be something that I am not seeing that would make that problematic.

gmalecha-at-skylabs (dismissed):

rocq-skylabs-brick/theories/lang/cpp/syntax/typed.v
pgiarrusso-sl: That's pre-existing tho.

pgiarrusso-sl: No objection to deleting `ext_tu`, but our chat leads to a bigger objection.

- Requirement: (non-dependent) typecheckers should be compositional; replacing an expression/statement with a type with another expression having the same type should preserve typing for the rest of the program, and we want this to be obvious by construction.
- Exposing `ext_values` breaks that property.
- We only need to check that the `ObjValue` has `okind = tCtor`, nothing else. Adding `ext_kind` here would be okay.
- It _seems_ `ext_tu` was introduced to enforce this property, but that's not documented, which is why contributors didn't preserve it.

rocq-skylabs-brick/theories/lang/cpp/syntax/typed.v
gmalecha-at-skylabs: Moved to https://github.com/SkyLabsAI/BRiCk/pull/262

rocq-skylabs-brick/theories/lang/cpp/syntax/typed.v
```diff
@@ -114,18 +113,22 @@ Module decltype.
       | None => throw ("failed to find global "%bs, n)
       end.
 
-    Definition value (n : name) : M ObjValue :=
-      let* osym := readerT.asks (fun '(tu, _, _, _) => tu.(ext_values) n) in
-      match osym with
-      | Some ov => mret ov
-      | None => throw ("failed to find global value "%bs, n)
+    Fixpoint map2 {T U V : Set} (f : T -> U -> M V) (ts : list T) (us : list U) : M (list V) :=
```
pgiarrusso-sl, line 116: No action needed but if you zip the two lists together, what's left is just 
`(f : TU -> M V) (tus : list TU) : M (list V)`, aka `traverse :: Applicative f => (a -> f b) -> t a -> f (t b)` with `F := M` and `t := list`.

pgiarrusso-sl (commented):
LGTM, I'd approve your changes and they passed CI.

gmalecha-at-skylabs (approved):
Mostly my code. Someone else should review as well.

PR #259, opened 2026-07-09
Make specification symbols opaque

gmalecha-at-skylabs:
Make specification symbols typeclass opaque

### DIFF
--- a/rocq-skylabs-cpp2v/src/ToCoq.cpp
+++ b/rocq-skylabs-cpp2v/src/ToCoq.cpp
@@ -359,7 +359,7 @@ void ToCoqConsumer::toCoqModule(clang::ASTContext *ctxt,
         End meta.
 
         Definition source := with_templates static_tu meta_tu.
-        #[deprecated]
+        #[warnings="-deprecated-syntactic-definition"]
         Abbreviation module := source.
         ```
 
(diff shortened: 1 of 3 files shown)


Janno:
Impressive speedup in reset_cpu.v :)

gmalecha-at-skylabs:
This seems promising, but the breakage is a bit subtle. It seems like we should potentially make the definitions that the typeclass instances unfold to `Hint Opaque` since we are sometimes seeing those.

rocq-skylabs-cpp2v/src/ToCoq.cpp
```diff
@@ -359,7 +359,7 @@ void ToCoqConsumer::toCoqModule(clang::ASTContext *ctxt,
         End meta.
 
         Definition source := with_templates static_tu meta_tu.
-        #[deprecated]
+        #[warnings="-deprecated-syntactic-definition"]
```
pgiarrusso-sl, line 362: I'm still confused about `-deprecated-syntactic-definition`, this change seems surprising to me.
```suggestion
```

rocq-skylabs-cpp2v/src/ToCoq.cpp
```diff
@@ -419,7 +419,7 @@ void ToCoqConsumer::toCoqModule(clang::ASTContext *ctxt,
 
         if (!interactive_.has_value()) {
             // NOTE: Backwards compatibility
-            fmt << "#[deprecated(note=\"use [source] instead.\")]" << fmt::line
+            fmt << "#[warnings=\"-deprecated-syntactic-definition\"]" << fmt::line
```
pgiarrusso-sl, line 422: Ditto.

rocq-skylabs-cpp2v/src/ToCoq.cpp
```diff
@@ -467,6 +467,8 @@ void ToCoqConsumer::toCoqModule(clang::ASTContext *ctxt,
             } else {
                 // NOTE: Backwards compatibility
                 print.output()
+                    << "#[warnings=\"-deprecated-syntactic-definition\"]"
```
pgiarrusso-sl, line 470: Ditto

PR #264, opened 2026-07-16
Resolution with default template parameters

### LINKED ISSUE #257: Support resolving names through default template parameters
The standard definition of `vector` looks like:

```c++
template<typename T, typename ALLOC = std::allocator<T>>
class vector
```

This allows users to write `std::vector<int>` in C++ but cpp2v does not track this data so users are required to write `std::vector<int, std::allocator<int>>`.

We should make resolution find this.

gmalecha-at-skylabs:
Closes https://github.com/SkyLabsAI/BRiCk/issues/257


rocq-skylabs-brick/theories/lang/cpp/syntax/translation_unit.v
```diff
@@ -473,11 +473,26 @@ Module template_alias.
     | Texprtype e => Texprtype (subst_expr xs e)
     end.
 
-  Definition same_template_base (actual candidate : name) : option (list temp_arg) :=
+  (** Match a template alias key against an actual name, collecting arguments
+      from nested instantiations such as [Outer<int>::Inner<bool>] against
+      [Outer<$T>::Inner<$U>]. Each instantiation level must have the same arity
+      so arguments from different scopes cannot be accidentally flattened
+      together. *)
+  Fixpoint same_template_base (actual candidate : name) : option (list temp_arg) :=
```
pgiarrusso-sl, line 481: This hunk is the entire change?

gmalecha-at-skylabs: I'm not sure what you mean by this. There are a lot of changes in this PR, most of them have to do with changes to cpp2v. Are you specifically talking about the comment?

pgiarrusso-sl: I was confused, no action needed.

To elaborate: The goal of the PR is to resolve `"std::vector<int>"%cpp_type` to `"std::vector<int, std::allocator<int>>"%cpp_type`, so I was expecting changes to Gallina code and got confused.

But Gallina already supports `Dtemplated_typedef`, so apparently we mostly change `cpp2v` to output more `Dtemplated_typedef` entries?

gmalecha-at-skylabs: That's correct.

pgiarrusso-sl (approved):

pgiarrusso-sl:
Replaced by #274.

PR #267, opened 2026-07-21
cpp.prog: Eliminate verbose stack traces in ill-typed terms

### LINKED ISSUE #248: Eliminate verbose stack traces in ill-typed terms from cpp.prog

gmalecha-at-skylabs:
Closes #248

### DIFF
--- a/rocq-skylabs-brick/plugin/src/cpp2v.ml
+++ b/rocq-skylabs-brick/plugin/src/cpp2v.ml
@@ -6,6 +6,12 @@
 open Procq.Constr
 open Stdarg
 
+(* Disable exception backtraces, overriding effects of the Base library.
+   Backtraces affect all Rocq user errors, not just anomalies. *)
+let _ =
+  let _ = Base.compare_option in
+  Printexc.record_backtrace false
+
 type report_level =
   | Ignore
   | Warn
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,3 +9,7 @@ Follow the conventions in [ROCQ_STYLE.md](ROCQ_STYLE.md).
 ## OCaml Style
 
 Always use `dune b @fmt` to re-format OCaml code according to the coding convention.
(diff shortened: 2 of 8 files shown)


rocq-skylabs-brick/plugin/src/cpp2v.ml
```diff
@@ -6,6 +6,12 @@
 open Procq.Constr
 open Stdarg
 
+(* Disable exception backtraces, overriding effects of the Base library.
+   Backtraces affect all Rocq user errors, not just anomalies. *)
+let _ =
+  let _ = Base.compare_option in
+  Printexc.record_backtrace false
+
```
rlepigre-skylabs-ai, line 14: I don't understand how this can help, is the `Base` library loaded in some other plugin that we use? If yes, shouldn't we do this fixing-up in that other plugin?

pgiarrusso-sl: We load `inline_tests`, and `inline_tests` loads `Base` — and this is the fix we figured out before in `fmdeps/auto/rocq-skylabs-auto-core/plugins/automation/auto_base.ml`. There isn't another SkyLabs plugin where we could add this code.

This was annoying us especially Gregory, so I suggested him copy-paste these lines for now; I can revisit in https://github.com/SkyLabsAI/BRiCk/pull/249.

pgiarrusso-sl (approved):

rocq-skylabs-brick/plugin/src/cpp2v.ml
gmalecha-at-skylabs: I'm okay if we don't get this merged. I think this is important to do in the future though since

1. This is basically all noise so it will fill agent context while providing almost no value.
2. The noise might not be stable between Rocq versions (which will impact supporting multiple Rocq versions if that is desired)

PR #268, opened 2026-07-21
A simple tool that checks elpi line numbers

gmalecha-at-skylabs:


gmalecha-at-skylabs:
We'll probably replace this with a patch to rocq-elpi.

gmalecha-at-skylabs:
Apparently no patch is needed. These annotations can simply be removed.

PR #273, opened 2026-07-24
Inline [function_type]

gmalecha-at-skylabs:


pgiarrusso-sl (approved):

gmalecha-at-skylabs:
Performance-wise, this doesn't seem like as much of a win as I expected. I don't think that it makes sense to merge this.

PR #274, opened 2026-07-24
Default template parameters in TUs

### LINKED ISSUE #257: Support resolving names through default template parameters

gmalecha-at-skylabs:
Beyond a few simplifications and bug fixes, this adds default template parameters to translation units.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/syntax/mcore.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/syntax/mcore.v
@@ -60,9 +60,6 @@ template file) and arguments.
 Definition temp_name : Set := name.
 
 Section tpreinst.
-  (* TODO: this type probably does not need to be parametric in [lang.t]
-     The only meaningful instantation is [lang.cpp]
-   *)
   Record tpreinst : Set := TPreInst {
     tpreinst_name : temp_name;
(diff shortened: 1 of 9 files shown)


rocq-skylabs-brick/theories/lang/cpp/syntax/mcore.v
```diff
@@ -116,7 +113,7 @@ Section template.
   Context {V : Type@{uV}}.
 
   Record template : Type@{uV} := Template {
-    template_params : list temp_param;
+    template_params : list (temp_param * option temp_arg);
```
gmalecha-at-skylabs, line 116: Unfusing `temp_param` would simplify some additional code.

pgiarrusso-sl: https://github.com/SkyLabsAI/BRiCk/issues/276

pgiarrusso-sl (approved):

PR #279, opened 2026-08-07
Fix name parsing/printing for user_defined_literals

pgiarrusso-sl:


rlepigre-skylabs-ai (approved):
Looks great to me. I guess you plan to remove the commit with the `_RocqProject` file?

PR #281, opened 2026-08-10
feat: `std::initializer_list`

jhaag-skylabs-ai:
> [!TIP]
>
> References:
> - https://eel.is/c++draft/dcl.init.list
> - https://en.cppreference.com/cpp/utility/initializer_list

Add basic support for `std::initializer_list` based on existing brace-initialization semantics.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,32 +15,10 @@ implicit construction of a `std::initializer_list<E>` from a braced-init-list
 (<http://eel.is/c++draft/dcl.init.list#5>).
 - New `Expr` constructor `Einitlist_std (backing : Expr) (t : type)`. Downstream
   exhaustive matches over `Expr` need a new case.
-- New **abstract** representation predicate `initializer_listR ety q arrayp n`
-  in `lang/cpp/logic/expr.v`. This is what specifications of
-  `std::initializer_list` should be written against. Its interface is the global
-  instances `initializer_listR_cfractional`, `initializer_listR_type_ptr` and
-  `initializer_listR_agree`, plus the derived `initializer_listR_ascfractional`.
-  Destruction is *not* part of it: that goes through the standard library's
-  destructor specification, which consumes the predicate directly.
-
-  It is abstract because `std::initializer_list` has no specified data members,
-  so every conforming representation is isomorphic to the (base pointer, element
-  count) pair the predicate carries. The rationale is stated in full at
-  `initializer_listR`; it is not repeated elsewhere.
-
-  `ety` is the class's **template argument**, not the backing array's element
-  type — <http://eel.is/c++draft/dcl.init.list#5> makes the array `const ety[n]`,
-  so the two differ by that `const`. See `std_initializer_list_element` in
-  `lang/cpp/syntax/types.v`.
-- New `Tstd_initializer_list ety` abbreviation and
-  `std_initializer_list_element` in `lang/cpp/syntax/types.v`, alongside the
-  existing `std_initializer_list` name.
-- `wp_init_initlist_std` grants `initializer_listR` for the backing array it
-  evaluates. It requires the class to be `std::initializer_list<E>` and the
-  backing array to be `const E[N]` for that same `E`.
-  `lang/cpp/syntax/typed.v` checks a weaker condition — it does not require a
-  known extent — so `cpp2v --check-types` rejects malformed nodes but accepting
-  a node does not by itself mean the rule applies to it.
+- `wp_init_initlist_std` reduces it to the `std::initializer_list` constructor
+  named by `std_initlist_ctor` (`lang/cpp/syntax/types.v`), following
+  `wp_init_binop_spaceship`. The representation of the resulting object belongs
+  to whoever specifies that constructor; see `brick-libcpp`.
 - The lifetime-extended case, e.g. `std::initializer_list<int> il = {1,2,3};`,
   is still unsupported because it requires scope-extruded temporaries.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,40 @@ of the release, and introduce a new heading.
 
 ## Since Last Release
 
+2026.07.29: Add support for clang's `CXXStdInitializerListExpr`, i.e. the
+implicit construction of a `std::initializer_list<E>` from a braced-init-list
+(<http://eel.is/c++draft/dcl.init.list#5>).
+- New `Expr` constructor `Einitlist_std (backing : Expr) (t : type)`. Downstream
+  exhaustive matches over `Expr` need a new case.
+- New **abstract** representation predicate `initializer_listR ety q arrayp n`
+  in `lang/cpp/logic/expr.v`. This is what specifications of
+  `std::initializer_list` should be written against. Its interface is the global
+  instances `initializer_listR_cfractional`, `initializer_listR_type_ptr` and
+  `initializer_listR_agree`, plus the derived `initializer_listR_ascfractional`.
+  Destruction is *not* part of it: that goes through the standard library's
+  destructor specification, which consumes the predicate directly.
+
+  It is abstract because `std::initializer_list` has no specified data members,
+  so every conforming representation is isomorphic to the (base pointer, element
+  count) pair the predicate carries. The rationale is stated in full at
+  `initializer_listR`; it is not repeated elsewhere.
+
+  `ety` is the class's **template argument**, not the backing array's element
+  type — <http://eel.is/c++draft/dcl.init.list#5> makes the array `const ety[n]`,
+  so the two differ by that `const`. See `std_initializer_list_element` in
+  `lang/cpp/syntax/types.v`.
+- New `Tstd_initializer_list ety` abbreviation and
+  `std_initializer_list_element` in `lang/cpp/syntax/types.v`, alongside the
+  existing `std_initializer_list` name.
+- `wp_init_initlist_std` grants `initializer_listR` for the backing array it
+  evaluates. It requires the class to be `std::initializer_list<E>` and the
+  backing array to be `const E[N]` for that same `E`.
+  `lang/cpp/syntax/typed.v` checks a weaker condition — it does not require a
+  known extent — so `cpp2v --check-types` rejects malformed nodes but accepting
+  a node does not by itself mean the rule applies to it.
+- The lifetime-extended case, e.g. `std::initializer_list<int> il = {1,2,3};`,
+  is still unsupported because it requires scope-extruded temporaries.
+
 2025.03.20: Change global prefix `bedrock.xxx` to `bluerock.xxx`
 - See `scripts/fm-refactorings/bedrock-bluerock.sh`
 
--- a/rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
@@ -29,6 +29,26 @@ Require Import skylabs.lang.cpp.logic.dispatch.
 Require Import skylabs.lang.cpp.logic.func.
 Require Import skylabs.iris.extra.bi.errors.
 
+(**
+[std_initlist_ctor_available tu cls aety] holds when [cls] declares the
+constructor that [wp_init_initlist_std] reduces [Einitlist_std] to.
+
+NOTE [Nctor] names carry their parameter types, so finding [std_initlist_ctor]
+under this name *is* finding the <<(const E*, size_t)>> signature; there is
+nothing further to compare. See [std_initlist_ctor] in ../syntax/types.v for why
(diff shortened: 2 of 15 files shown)


rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -49,6 +49,82 @@ Module Type Expr.
...
+
+  NOTE the interface below is what specifications of <<std::initializer_list>>
+  are written against, so it is the expensive thing to change. Should we ever
+  want to *prove* a standard library rather than axiomatize it, this predicate
+  can be given a definition in terms of that library's fields, at which point
+  these instances become theorems and no specification stated in terms of them
+  needs to change.
+  *)
+  Parameter initializer_listR : forall `{Σ : cpp_logic, σ : genv}
+    (ety : type) (q : cQp.t) (arrayp : ptr) (n : N), Rep.
```
pgiarrusso-sl, line 101: The design here is not obvious, but for `wp_init_binop_spaceship`, me and Gregory chose to leave the `Rep` downstream in the library specs (`brick-libcpp` and `rocq-skylabs-cpp-stdlib`).

To follow that pattern, `wp_init_initlist_std` would turn `wp_init ty base (Einitlist_std backing ty) Q` into the `wp_init` of some `std::initializer_list` constructor, after checking that the new syntax node is well-typed.

rocq-skylabs-brick/theories/lang/cpp/syntax/types.v
```diff
@@ -1271,6 +1271,91 @@ Definition Tmember_func (ty : exprtype) (fty : functype) : functype :=
...
+We spell the name out rather than using the <<cpp_name>> notation because that
+notation's parser depends on this file. [TEST_std_initializer_list_name] in
+../logic/expr.v checks the two agree.
+
+NOTE [std_initializer_list] and [Tstd_initializer_list] are [Abbreviation]s, not
+[Definition]s, so that a goal mentioning the class contains the name itself. An
+opaque constant could not be matched against the name that specifications use,
+and a transparent one is rejected by the <<sl-transparent-constants>> check when
+it appears under [type_ptr]. They sit outside [with_lang] because a
+section-local [Abbreviation] would be discharged at [End with_lang].
```
pgiarrusso-sl, line 1291: True but too much IMO:

```suggestion

NOTE [std_initializer_list] and [Tstd_initializer_list] must be [Abbreviation]s,
not hidden behind a [Definition]s, since ASTs will expand them.
```

or
```suggestion
```

rocq-skylabs-brick/theories/lang/cpp/syntax/types.v
```diff
@@ -1271,6 +1271,91 @@ Definition Tmember_func (ty : exprtype) (fty : functype) : functype :=
...
+
+The class from <https://eel.is/c++draft/dcl.init.list#5>.
+
+NOTE this name is fixed by the standard and is never versioned: libc++ declares
+<<initializer_list>> *outside* its inline versioning namespace precisely because
+the compiler must be able to find <<std::initializer_list>>.
+
+We spell the name out rather than using the <<cpp_name>> notation because that
+notation's parser depends on this file. [TEST_std_initializer_list_name] in
+../logic/expr.v checks the two agree.
```
pgiarrusso-sl, line 1284: We could move this downstream TBH — and at least `std_initializer_list` could become `"std::initializer_list"%cpp_name` (with the scope annotation usually omitted).

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -2136,3 +2260,17 @@ Declare Module E : Expr.
...
+  std_initializer_list = "std::initializer_list"%cpp_name := eq_refl.
+
+(** Follows from [initializer_listR_cfractional]; it is what lets the IPM see
+    [initializer_listR] as fractional when solving for a fraction. It cannot be
+    declared alongside the other instances in [Expr], since it has a proof. *)
+#[global] Instance initializer_listR_ascfractional `{Σ : cpp_logic, σ : genv}
+    (ety : type) (arrayp : ptr) (n : N) (q : cQp.t) :
+  AsCFractional (initializer_listR ety q arrayp n)
+    (fun q => initializer_listR ety q arrayp n) q.
+Proof. solve_as_cfrac. Qed.
```
pgiarrusso-sl, line 2276: > It cannot be
    declared alongside the other instances in [Expr], since it has a proof.

Slop. Proofs in there are just fine.

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -1959,6 +2035,54 @@ Module Type Expr.
...
+        builds the <<std::initializer_list>> object at [base] referring to it.
+
+        The resulting object is described by [initializer_listR] (declared at
+        the top of this file), which is abstract; see the rationale there. In
+        particular this rule commits to no field layout, and so to none of
+        clang's choices about one.
+
+        NOTE the backing array is a temporary, so its storage (and the
+        corresponding [FreeTemps]) come from the [Ematerialize_temp] that
+        [backing] wraps. Consequently this rule only covers backing arrays whose
```
pgiarrusso-sl, line 2051: Do we have an end-to-end test of this rule with `Import linearity.`? (Checks) Not yet, based on 
https://github.com/SkyLabsAI/brick-libcpp/pull/112/changes#diff-8db47bc146b66171a3df36fc098e5987a7518ce7d6cd8c60212d03d98fb4bb88, hopefully it should be fixable.

I wonder if we need to free `backing` ourselves  (say `Q (delete ? backing >*> free)` instead of `Q free` or such) or if `Ematerialize_temp` handles it — but let's not waste words on the question, the test will save us that discussion.

pgiarrusso-sl (changes requested):
I think this PR looks pretty good, but I'd prefer moving the initializer_list `Rep` downstream.

Some of the comments are too verbose even for me.

CHANGES.md
```diff
@@ -10,6 +10,40 @@ of the release, and introduce a new heading.
...
+  `std_initializer_list_element` in `lang/cpp/syntax/types.v`, alongside the
+  existing `std_initializer_list` name.
+- `wp_init_initlist_std` grants `initializer_listR` for the backing array it
+  evaluates. It requires the class to be `std::initializer_list<E>` and the
+  backing array to be `const E[N]` for that same `E`.
+  `lang/cpp/syntax/typed.v` checks a weaker condition — it does not require a
+  known extent — so `cpp2v --check-types` rejects malformed nodes but accepting
+  a node does not by itself mean the rule applies to it.
+- The lifetime-extended case, e.g. `std::initializer_list<int> il = {1,2,3};`,
+  is still unsupported because it requires scope-extruded temporaries.
```
pgiarrusso-sl, line 45: This is definitely too verbose.

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
pgiarrusso-sl: I forgot to add: while that rule is "more expensive" for clients to use, something like `wp_init_initlist_std` can be verified as a _hint_ on top.

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -49,6 +49,82 @@ Module Type Expr.
...
+        array. *)
+    #[global] Declare Instance initializer_listR_type_ptr ety q arrayp n :
+      Observe (type_ptrR (Tstd_initializer_list ety))
+              (initializer_listR ety q arrayp n).
+
+    (** Two fractions of the same object refer to the same backing array. *)
+    #[global] Declare Instance initializer_listR_agree ety q q' arrayp arrayp' n n' :
+      Observe2 [| arrayp = arrayp' /\ n = n' |]
+        (initializer_listR ety q  arrayp  n)
+        (initializer_listR ety q' arrayp' n').
```
pgiarrusso-sl, line 125: Preference: I still prefer a single model (pair or record) — then this rule would be `Agree1` and we'd have slightly more standard support for it.
```
Parameter initializer_listR : forall `{Σ : cpp_logic, σ : genv}
    (ety : type) (q : cQp.t) (model : ptr * N), Rep.
```

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
jhaag-skylabs-ai: Sounds good, I'll move to this approach.

jhaag-skylabs-ai: @pgiarrusso-sl actually, the standard only prescribes the default constructor of an empty `std::initializer_list`, cf. https://eel.is/c++draft/support.initlist. IIUC, other constructors are implementation defined. Indeed, it seems clang/gcc use `initializer_list(const E* base, size_t size)` while MSVC uses `initializer_list(const E* begin, const E* end)`.

Do you still think that we should bake the (current) clang/gcc signature into BRiCk by producing a concrete `wp_init` with the corresponding `initializer_list(const E*, size_t)` signature?

pgiarrusso-sl: That's a very good question! 

But I think we should be okay with specializing to clang/gcc anyway, we do it in other places. This approach still lets us write a sufficient WP rule and define `initializer_listR` downstream.

- The axiom side conditions should check that "`initializer_list(const E* base, size_t size)` is available".
- And if we ever support MSVC, we could write the alternative rule — and it could test for the other overload and/or for the compiler flavor.

pgiarrusso-sl: Feel free to extend the documentation about limitations, but to be clear, I'll maintain this is a technicality. 
clang passes `(base, size)`, MSVC passes `(base, base + size)` with a slightly different spec. Unless either compiler has a bug, those two APIs are equivalent, and will validate the same code.

jhaag-skylabs-ai: Sounds like a plan. I'll ping you again once your feedback is incorporated and the PRs are ready for another review.

rocq-skylabs-brick/theories/lang/cpp/syntax/types.v
jhaag-skylabs-ai: For now I've kept this in BRiCk since the standard defines the name itself. Resolving, but let me know if you want me to downstream it.

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
jhaag-skylabs-ai: This is now tested w/`Import linearity.` in the corresponding [`brick-libcpp` PR](https://github.com/SkyLabsAI/brick-libcpp/pull/112). It does seem that `Ematerialize_temp` handles the backing array cleanup correctly.

CHANGES.md
jhaag-skylabs-ai: I trimmed essentially all of this text.

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
jhaag-skylabs-ai: The `brick-libcpp` PR bundles them in a record instead of a pair.

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -1959,6 +1979,40 @@ Module Type Expr.
...
+        <<std::initializer_list<int> il = {1,2,3};>>, additionally requires
+        scope-extruded temporaries, which BRiCk does not yet have; cpp2v emits
+        [Eunsupported] for those. *)
+    Axiom wp_init_initlist_std : forall cls (base : ptr) cv ty backing aety (n : N) Q,
+        decompose_type ty = (cv, Tnamed cls) ->
+        drop_qualifiers (drop_reference (type_of backing)) = Tarray aety n ->
+        std_initlist_ctor_available tu cls aety = true ->
+        wp_init ty base
+          (Econstructor (std_initlist_ctor cls aety)
+             [Ecast Carray2ptr backing; Eint (Z.of_N n) Tsize_t] ty) Q
```
pgiarrusso-sl: Based on `wp_init_binop_spaceship` and discussions with Gregory back then, we should add some typechecking — `decltype.of_expr` is currently trivial but it might become stricter in the future.

```suggestion
        let e := (Econstructor (std_initlist_ctor cls aety)
             [Ecast Carray2ptr backing; Eint (Z.of_N n) Tsize_t] ty) in
        [| decltype.of_expr e = Some ty |] ∗
        wp_init ty base e Q
```

pgiarrusso-sl (approved):

PR #285, opened 2026-08-15
Canonical free_temps: sketch interface and consumers

pgiarrusso-sl:
Not connected to anything, but sketched by Claude — initially without access to a build test.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
@@ -261,26 +261,11 @@ Section interp_nf.
     Proper (pointwise_relation _ (⊣⊢) ==> pointwise_relation _ (⊣⊢) ==> eq ==> (⊣⊢)) bi_par.
   Proof. solve_proper. Qed.
 
-  Lemma bi_par_comm_1 (K1 K2 K3 : mpred -> mpred) (Q : mpred) :
-    bi_par K1 (bi_par K2 K3) Q |-- bi_par K2 (bi_par K1 K3) Q.
+  Lemma bi_par_comm F G Q : bi_par F G Q ⊣⊢ bi_par G F Q.
   Proof.
-    rewrite /bi_par.
-    iIntros ">(%Q1 & %Q2 & K1 & >(%Q0 & %Q3 & K2 & K3 & W1) & W2) !>".
-    iExists Q0; iFrame "K2".
-    iExists (Q1 ** Q3).
-    iSplitR "W1 W2". {
-      iIntros "!>".
-      iExists Q1, Q3; iFrame. by iIntros; iFrame.
-    }
-    iIntros "(? & Q1 & ?)".
-    iApply ("W2" with "[>- $Q1]").
-    iApply "W1"; iFrame.
-  Qed.
-
-  Lemma bi_par_comm (K1 K2 K3 : mpred -> mpred) (Q : mpred) :
-    bi_par K1 (bi_par K2 K3) Q -|- bi_par K2 (bi_par K1 K3) Q.
-  Proof.
-    iSplit; iApply bi_par_comm_1.
+    rewrite /bi_par; iSplit; iIntros ">(%Qx & %Qy & HF & HG & Hw) !>";
+      iExists Qy, Qx; iFrame "HF HG"; iIntros "[H1 H2]";
+      by iApply ("Hw" with "[$]").
   Qed.
 
   Lemma bi_par_assoc K1 K2 K3 Q :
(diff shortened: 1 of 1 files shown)


rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
```diff
@@ -0,0 +1,727 @@
+(*
+ * Copyright (c) 2026 SkyLabs AI, Inc.
+ * This software is distributed under the terms of the BedRock Open-Source License.
+ * See the LICENSE-BedRock file in the repository root for details.
+ *)
+
+(** * Alternative [FreeTemps] / [interp]: abstract type, propositional laws.
+
+    SKETCH. Never compiled. All proofs are [Admitted] with candidate scripts
```
pgiarrusso-sl, line 9: Now this builds fine.

rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
```diff
@@ -0,0 +1,727 @@
...
+
+
+(** ** The semantic eliminator *)
+Implicit Types (tu : translation_unit).
+
+Section interp_nf.
+  Context `{Σ : cpp_logic} {σ : genv}.
+  Implicit Types (Q : epred).
+
+  Definition bi_par K1 K2 Q : mpred :=
```
pgiarrusso-sl, line 256: TODO lift.

rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
```diff
@@ -0,0 +1,727 @@
...
+Notation interp_pars tu xs :=
+  (interp_nf tu (nf_par xs)).
+
+
+(** ** [Shift] *)
+
+Section shift.
+  Context `{Σ : cpp_logic}.
+
+  Definition Shift (F : epred -> mpred) : Prop :=
```
pgiarrusso-sl, line 361: TODO lift, probably as typeclasses.

rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
```diff
@@ -0,0 +1,727 @@
...
+
+      There is no obligation from the law itself. The obligation is
+      [interp_par_eq] instantiated at [f := FreeTemps.id], where the LHS
+      collapses to [interp tu g Q].  Both directions go through with NO
+      affinity assumption: the [⊢] direction picks [Qi := emp] and
+      PRODUCES [emp] rather than discarding a resource. *)
+  Lemma par_id_obligation g Q :
+    interp tu g Q ⊣⊢
+      bi_par (fupd top top) (interp tu g) Q.
+  Proof. by rewrite /interp -{1}bi_par_fupd_interp_nf. Qed.
```
pgiarrusso-sl, line 512: Drop, probably, this is just a sanity check exercise.

rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
```diff
@@ -0,0 +1,727 @@
...
+    (Q -* Q') |-- interp tu f Q -* interp tu f Q'.
+  Proof. apply interp_nf_frame. Qed.
+
+  Lemma interp_shift f Q :
+    (|={top}=> interp tu f (|={top}=> Q)) |-- interp tu f Q.
+  Proof. apply interp_nf_shift. Qed.
+End interp_equations.
+
+
+(** ** The automation eliminator
```
pgiarrusso-sl, line 624: Does not belong here.

rocq-skylabs-brick/theories/lang/cpp/logic/free_temps_destroy_alt.v
```diff
@@ -0,0 +1,727 @@
...
+      [Proper] and nothing needs to carry a canonicity proof. *)
+  #[global] Declare Instance seq_assoc : Assoc   (=) seq.
+  #[global] Declare Instance seq_id_l  : LeftId  (=) id seq.
+  #[global] Declare Instance seq_id_r  : RightId (=) id seq.
+  #[global] Declare Instance par_assoc : Assoc   (=) par.
+  #[global] Declare Instance par_comm  : Comm    (=) par.
+  #[global] Declare Instance par_id_l  : LeftId  (=) id par.
+
+  Axiom delete_ref : forall ty p, delete (Tref ty) p = delete (Trv_ref ty) p.
+End FREE_TEMPS.
```
pgiarrusso-sl, line 92: Open question — how do we implement this?

Janno: [Rewrite rules](https://rocq-prover.org/doc/V9.2.0/refman/addendum/rewrite-rules.html). They are pretty much made for stuff like this

pgiarrusso-sl: Rewrite rules can cause anomalies, I am not sure think we should enable them anywhere. They're the only way to add _definitional_ equalities, but I don't think we need that here.

I'd rather use one of the known implementation techniques for quotients in Rocq, like for pointers, but I don't think that should block the merging.

# More on rewrite rules

From your link:

> Rewrite rules are highly experimental.

> In particular, ill-typed rewrite rules will lead to mistyped expressions, and manipulating these will most often result in inconsistencies and anomalies.

AFAICT, nowadays they're most useful to type theorists prototyping new type theories who plan to prove on papers the rules are reasonable, but want to extend definitional equality without writing their own theorem prover.

Janno: Oh, I think I misunderstood your question. I assumed getting a model that would give us definitional equality here would be either impossible or infeasible. That's why I suggested rewrite rules. I agree that they should be the absolute last resort and they definitely have a bunch of problems.

PR #286, opened 2026-08-18
Extend parser for names in anonymous namespaces

pgiarrusso-sl:
Also add tests to the existing printer support.

This is really minimal, but agents managed to exploit the minimal support already, so we're just making it more convenient.

Proper semantics for internal linkage remains unsupported.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/syntax/name_notation/parser.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/syntax/name_notation/parser.v
@@ -579,7 +579,7 @@ Module internal.
         in
         commit (keyword "operator") operator
         $ commit (op_token "~") (fun _ => Dtor <$> ident)
-        $ commit (exact "(anon)") (fun _ => mret Anonymous)
+        $ commit (exact "(anonymous namespace)") (fun _ => mret Anonymous)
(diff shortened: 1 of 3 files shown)


pgiarrusso-sl:
Found based on the code that @skylabs-Usama shared.

cc @rlepigre, as update from yesterday. I think everything we discussed remains true.

pgiarrusso-sl:
@rlepigre also, if this goes in, `pm` would have to print the names in the same way. Unsure how easy that is!

rlepigre-skylabs-ai (approved):
Looks good to me! I'll see how easy it is to adapt `pm`, but I'm not too worried.

pgiarrusso-sl:
@rlepigre also, if this goes in, `pm` would have to print the names in the same way. Unsure how easy that is!

rlepigre-skylabs-ai (changes requested):
Actually, it seems that Clang uses `(anonymous namespace)::internal()` instead of `(anon)::internal()`. Do you think we should match that?

rlepigre-skylabs-ai:
For instance:
```sh-session
$ llvm-cxxfilt _ZN12_GLOBAL__N_18internalEv
(anonymous namespace)::internal()
```
So this would then be immediately compatible with our other tooling,

rlepigre-skylabs-ai (approved):

pgiarrusso-sl:
Thanks for completing this one!

PR #296, opened 2026-08-26
Add opaque location information 

Janno:


rocq-skylabs-brick/theories/lang/cpp/logic/heap_pred/tptsto.v
```diff
@@ -30,6 +30,39 @@ Section tptstoR.
 
   (** [tptstoR] *)
 
+  Axiom tptstoR_loc_info : forall li ty q v,
+      tptstoR (TLocInfo li ty) q v = tptstoR ty q v.
+
+  Axiom tptsto_fuzzyR_loc_info : forall li ty q v,
+      tptsto_fuzzyR (TLocInfo li ty) q v = tptsto_fuzzyR ty q v.
```
pgiarrusso-sl: Stating equalities here seems a bit too strong when R ** emp <> R.
And I'd rather prove the bi-entailments from axioms on tptsto.

rocq-skylabs-brick/theories/lang/cpp/logic/destroy.v
```diff
@@ -150,6 +150,10 @@ Section prim.
   Context `{Σ : cpp_logic, σ : genv}.
   Implicit Types (Q : epred).
 
+  Axiom wp_destroy_prim_loc_info : forall tu cv li ty (this : ptr) Q,
+    wp_destroy_prim tu cv (TLocInfo li ty) this Q -|-
+    wp_destroy_prim tu cv ty this Q.
```
pgiarrusso-sl: Prove this by unfolding wp_destroy_prim.

Axioms about definitions are a code smell.

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -208,8 +208,9 @@ Module Type Expr.
        type is not a reference, then the pointer is returned *after* it is
        checked to be strictly valid.
      *)
-    Definition read_decl (p : ptr) (t : type) (Q : ptr -> mpred) :=
-      match drop_qualifiers t with
+    Definition read_decl (p : ptr) (t : type) (Q : ptr -> mpred) : mpred :=
+      let t := drop_loc_info (drop_qualifiers t) in
```
pgiarrusso-sl: Is this sequence legal?

rocq-skylabs-brick/theories/lang/cpp/logic/heap_pred/tptsto.v
Janno: Have another look please. We now have an axiom directly on `tptsto`: https://github.com/SkyLabsAI/BRiCk/pull/296/changes#diff-c8979bfafcf3fbd7246381cb27bf71ae7758d16df806d5ec80f9f4f7b04f5020R295

Janno:
New axioms in `BRiCk` (produced by `git diff origin/main HEAD | rg '^\+' | rg 'Axiom' -A 3` and some manual postprocessing)
```coq
Axiom has_type_loc_info : ∀ li ty v,
  has_type v (TLocInfo li ty) -|- has_type v ty.

Axiom reference_to_loc_info : forall li ty p,
  reference_to (TLocInfo li ty) p -|- reference_to ty p.

Axiom tptsto_loc_info : forall li ty q p v,
  tptsto (TLocInfo li ty) q p v -|- tptsto ty q p v.

Axiom wp_lval_loc_info : forall {σ : genv} tu ρ li e Q,
    wp_lval tu ρ (ELocInfo li e) Q -|- wp_lval tu ρ e Q.

Axiom wp_init_expr_loc_info : forall {σ : genv} tu ρ ty p li e Q,
    wp_init tu ρ ty p (ELocInfo li e) Q -|- wp_init tu ρ ty p e Q.

Axiom wp_operand_loc_info : forall {σ : genv} tu ρ li e Q,
    wp_operand tu ρ (ELocInfo li e) Q -|- wp_operand tu ρ e Q.

Axiom wp_xval_loc_info : forall {σ : genv} tu ρ li e Q,
    wp_xval tu ρ (ELocInfo li e) Q -|- wp_xval tu ρ e Q.

Axiom wp_loc_info : forall σ tu ρ li s Q,
    wp (resolve:=σ) tu ρ (SLocInfo li s) Q -|- wp tu ρ s Q.

Axiom align_of_loc_info : ∀ li t,
    align_of (TLocInfo li t) = align_of t.

Axiom has_type_prop_loc_info : forall v li ty,
    has_type_prop v (TLocInfo li ty) <-> has_type_prop v ty.
```

rlepigre-skylabs-ai (approved):
This looks good to me, but it might be good for @pgiarrusso-sl to take another look.

pgiarrusso-sl (changes requested):
Do we have erasure lemmas for `type_ptr`? I don't see an axiom about it, and I'm not sure it's provable?

Thanks for adding `has_type_loc_info`, but what do we do with `has_type (Tarray (TLocInfo li ty))`? I think this MR needs an `erase_loc_info` (modeled after `erase_qualifiers`). EDIT: it's called `erase_type`, but it seems surprising we don't see it in the axioms — does it appear in derived lemmas?

And then, `same_address_eq_type_ptr` will have to require `erase_qualifiers (erase_loc_info ty) <> Tuchar` (building on https://github.com/SkyLabsAI/BRiCk/pull/299). It's possible other axioms and code locations need extra calls to `erase_loc_info` to respect the new type equivalence that `TLocInfo` induces.

And quite possibly, instead of separate `drop_loc_info` and `drop_qualifiers`, we want a fused operation — composing them seems slower. Maybe that can wait for clearer performance numbers. OTOH, it's possible _all_ callers to `drop_qualifiers` need `erase_loc_info`.

rocq-skylabs-brick/theories/lang/cpp/syntax/core.v
```diff
@@ -558,11 +568,350 @@ Inductive temp_param : Set :=
...
+#[global] Arguments drop_atomic_name_loc_info !_ / : simpl nomatch, assert.
+#[global] Arguments drop_name_loc_info !_ / : simpl nomatch, assert.
+#[global] Arguments drop_temp_arg_loc_info !_ / : simpl nomatch, assert.
+#[global] Arguments drop_loc_info !_ / : simpl nomatch, assert.
+#[global] Arguments drop_expr_loc_info !_ / : simpl nomatch, assert.
+#[global] Arguments drop_stmt_loc_info !_ / : simpl nomatch, assert.
+#[global] Arguments drop_var_decl_loc_info !_ / : simpl nomatch, assert.
+#[global] Arguments drop_binding_decl_loc_info !_ / : simpl nomatch, assert.
+
+Fixpoint erase_atomic_name (an : atomic_name) {struct an} : atomic_name :=
```
pgiarrusso-sl: Naming question — `erase_type` only erases `*LocInfo` not qualifiers, so it seems potentially misleading. Especially because ~~`drop_type_loc_info`~~ `drop_loc_info` does mention `loc_info` in its name.

(I found out because I was hoping for something that erases everything, and the name seemed encouraging).

rocq-skylabs-brick/theories/lang/cpp/syntax/translation_unit.v
```diff
@@ -367,13 +367,14 @@ Module template_alias.
...
+  Fixpoint subst_expr (xs : env) (e : Expr) : Expr :=
     match e with
     | Eparam id =>
         match lookup_value xs id with
         | Some e => e
         | None => e
         end
+    | ELocInfo li e => ELocInfo li (subst_expr xs e)
     | _ => e
     end.
```
pgiarrusso-sl, line 379: I can't quite blame you, but "recurse _sometimes_ under `Expr` constructors" seems a surprising semantics here 💀 — could it be a bug?

Janno: No, I don't think this is a bug. The fact that a parameter has location info attached to it should not stop us from substituting it, should it?

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
Janno: In what sense of the word "legal"? I think `drop_loc_info` is not going to do anything here given that `drop_qualifiers` cannot produce a `TLocInfo` node.

rocq-skylabs-brick/theories/lang/cpp/syntax/core.v
Janno: Yeah, that's not a great name. I can add `_loc_info` for now to fix the confusing part. I don't feel confident adding functions that drop multiple things because I am not sure what to drop from where.

Janno (commented):
> EDIT: it's called erase_type, but it seems surprising we don't see it in the axioms — does it appear in derived lemmas?

I prompted the model to avoid full location erasure whenever possible so I wouldn't have to carve out a gazillion exceptions in all the places where the automation needs to have access to it.

> It's possible other axioms and code locations need extra calls to `erase_loc_info` to respect the new type equivalence that `TLocInfo` induces.

> And quite possibly, instead of separate drop_loc_info and drop_qualifiers, we want a fused operation — composing them seems slower. Maybe that can wait for clearer performance numbers. OTOH, it's possible all callers to drop_qualifiers need erase_loc_info.

Is there a setoid that we want to define to prove `Proper` for our constructions? I am not advocating to actually use the `Proper` facts, just to write them down to document which parts of BRiCk care about qualified types (and, now, location info).

rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
pgiarrusso-sl: > `drop_qualifiers` cannot produce a `TLocInfo` node.

but it can, right?
```
Fixpoint drop_qualifiers (t : type) : type :=
  match t with
  | Tqualified _ t => drop_qualifiers t
  | TLocInfo li t => TLocInfo li (drop_qualifiers t)
  | _ => t
  end.
```

My initial question is wrong — I was wondering if `drop_loc_info (drop_qualifiers t)` was actually dropping both, or you needed `drop_qualifiers (drop_loc_info t)` — but the actual implementation means they commute.

rocq-skylabs-brick/theories/lang/cpp/syntax/translation_unit.v
pgiarrusso-sl: I'm just skeptical the existing `subst_expr` function can possibly be correct.

pgiarrusso-sl:
> Is there a setoid that we want to define to prove Proper for our constructions? I am not advocating to actually use the Proper facts, just to write them down to document which parts of BRiCk care about qualified types (and, now, location info).

We probably have _multiple_ setoids, but yes, it might easier to define congruences than canonicalizers correctly.
Intuitively, everything should respect `erase_loc_info`, and I don't think the axioms Rodolphe showed are enough for that.

IMHO `drop_qualifiers` should at least `drop_loc_info` — unless you can show me an example where you're calling `drop_qualifiers` without calling `drop_loc_info` and is correct.

Similarly, `erase_qualifiers` should `erase_loc_info`.

pgiarrusso-sl:
Proposal:
- drop `TLocInfo` from this PR. That drops most location information from types. Expression in types remain and still have loc info.
- later, in the cpp2v changes, maybe omit loc info from expressions embedded in types — and figure out any problems as they appear?

rocq-skylabs-brick/theories/lang/cpp/syntax/translation_unit.v
pgiarrusso-sl: Resolution: add a comment "TODO figure out if this shallow substitution needs to be a real substitution"?

rocq-skylabs-brick/theories/lang/cpp/syntax/core.v
pgiarrusso-sl: Options: `erase_foo_li`, or namespacing in `LocInfo`, or overloading...

Janno:
Meeting notes:
- [ ] Use `erase_loc_info` in `GlobDecl_le`, `ObjValue_le`. Not a problem yet because no TUs have loc info yet so the inclusion still works. This will change with the follow-up branch that actually emits the information in cpp2v.
- [ ] Add `LocInfo` namespace module that contains erasure and drop functions. (Possibly use typeclasses to present unified functions for all syntactic kinds.)

Issue #310, opened 2026-09-06
Support raw representation for aggregates


pgiarrusso-sl:
We're missing principles for the raw representation of `structR`, analogous to `primR_to_rawsR`. Something like:
```
Axiom structR_to_anyRs : ∀ `{Σ : cpp_logic, σ : genv} ty q,
  structR ty q -|-
  type_ptrR ty ** Exists offs, [*list] off ∈ offs, anyR Tbyte q ** [| < state that offs only covers holes in the structure> |]
```

`σ : genv` has layout informations for `ty`: any byte in `ty`'s representation that isn't in a field representation belongs to `structR`'s representation.

We can derive rules that use translation units satisfying `genv_compat`.

pgiarrusso-sl:
~~That wasn't correct — we have `struct_to_raw`.~~

PR #311, opened 2026-09-07
Disable unsound [struct_to_raw], and other fixes (WIP)

pgiarrusso-sl:
`struct_to_raw` gives a base subobject the raw representation of a **complete** object of the base's type — `size_of` bytes rather than the bytes the base subobject actually occupies. This is unsound, and the everyday witness is the empty base optimization, not some exotic tail-padding corner:

```cpp
struct E { };                        // POD, size_of = 1, occupies 0 bytes as a base
struct D : E { unsigned char c; };   // standard-layout, size_of = 1, c at offset 0
```

Applied at `E` — no bases, no fields, `s_layout = POD`, so the side condition holds — the rule degenerates to

```coq
structR "E" q -|- type_ptrR (Tnamed "E") ** Exists rs, rawsR q rs ** [| raw_bytes_of_struct σ "E" ∅ rs |]
```

and `raw_bytes_of_struct_wf_size` forces `length rs = s_size E = 1`. So the *empty* base subobject owns a byte. Inside `D` that subobject sits at offset 0 — and so does `D::c`, whose byte is reachable through `primR_to_rawsR` and `eval_o_field` (`D` is standard-layout, so that side condition holds too). Two full-fraction claims on one byte give

```coq
Lemma ebo_False (p : ptr) : p |-> anyR (Tnamed "D") 1$m |-- False.
```

i.e. the representation predicate of an ordinary C++ class is unsatisfiable, and the logic is vacuous for any program containing an empty base.

This was confirmed by a machine-checked Rocq derivation (using `cpp.prog`, so the layout is clang's, not hand-written). `Print Assumptions` on it names, besides ordinary infrastructure: `struct_to_raw`, `raw_bytes_of_struct(_wf_size)`, `tptsto_ptr_congP_transport`, `eval_o_field`, and `eval_o_base`.

## Why the existing guards don't help

- The `s_layout ∈ [POD;Standard]` guard is satisfied: `E` is POD and `D` is standard-layout.
- A per-class layout-disjointness check on `E` would not help either — `E` has no members to be disjoint from. The offending layout belongs to `D`, while the over-claiming rule is applied at `E`.

## Ordering constraint on `eval_o_base`

The derivation needs one axiom that does not exist yet: `eval_o_base`, the base-class analogue of `eval_o_field` in `lang/cpp/semantics/ptrs.v`, with the same POD/Standard guard. Every concrete pointer model already computes exactly it (`simple_pointers_utils.o_base_off := parent_offset σ derived base`). So the bug is **latent in the current axiom set**: the only thing making it unreachable today is that no interface-level rule can evaluate a base offset.

Consequently `eval_o_base` must not be added until this rule is fixed — doing so first would make the axiom set outright inconsistent.

## Why comment out rather than weaken

`grep -rn "struct_to_raw" fmdeps/ bluerock/` returns no hits outside the declaration itself. With nothing to migrate, disabling it is the smallest step that removes the unsoundness, and it keeps the statement and the analysis in the tree for whoever writes the fix.

The fix, tracked by #310, is for a base subobject's extent to be the class's non-virtual size (0 for an empty class) rather than `size_of`, so the per-base entries in `rss` cover that instead.

## Validation

- `make -j8 stage1` — clean.
- `dune build @fmdeps/BRiCk/rocq-skylabs-brick/all` — clean; `layout.v` compiles with the axiom commented out.
- `dune build @vendored/install @fmdeps/all @fmdeps/runtest` — compiled the whole downstream Rocq graph (all of `rocq-skylabs-auto-cpp`, `rocq-skylabs-cpp-stdlib`, `rocq-brick-spec-strength`, `auto-docs`, `skylabs-fm/scaffold`) with **zero Rocq errors**, but did **not** finish clean on the machine used: 44 targets fail for a pre-existing environmental reason (no libstdc++ headers available to `cpp2v`; `CPP2V_DOCKER_ENABLED` unset), plus some macOS/AArch64 cram diffs (BSD `wc -l` padding, ARM `__Bfloat16x*_t` typedefs).

  This was confirmed independent of the patch by reproducing one of the failures (`fmdeps/auto/rocq-skylabs-auto-cpp/tests/destructuring/test_pair.vo`) on an unpatched checkout, and by checking that `clang++ -std=c++20 -stdlib=libstdc++ -fsyntax-only` fails standalone on a bare `#include <utility>`. CI should be the judge of a fully clean run.

Draft, pending review of the argument above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/logic/layout.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/logic/layout.v
@@ -26,7 +26,45 @@ Section with_Σ.
   (** Convert a <<struct>> to its raw representation.
    Justified by the concept of object representation.
    https://eel.is/c++draft/basic.types.general#def:representation,object
-   *)
+
+   DISABLED: this rule is unsound, because it gives a base subobject the raw
+   representation of a *complete* object of the base's type, i.e. [size_of]
+   bytes rather than the bytes the base subobject actually occupies. The
+   everyday witness is the empty base optimization:
+
+   <<
+     struct E { };                  // POD, size_of = 1, occupies 0 bytes as a base
+     struct D : E { unsigned char c; };   // standard-layout, size_of = 1, [c] at offset 0
+   >>
+
+   Applied at [E] (whose [s_layout] is [POD], so the side condition holds), the
+   rule reads [structR "E" q -|- type_ptrR ** Exists rs, rawsR q rs ** ...],
+   and [raw_bytes_of_struct_wf_size] forces [length rs = s_size E = 1]. So the
+   *empty* base subobject owns a byte -- the same byte as [D::c], which is
+   reachable through [primR_to_rawsR] and [eval_o_field] ([D] is
+   standard-layout, so that side condition holds too). Both at fraction 1
+   gives [p |-> anyR (Tnamed "D") 1$m |-- False]: the representation predicate
+   of an ordinary class is unsatisfiable, so the logic is vacuous for any
+   program containing an empty base.
+
+   Note the [s_layout ∈ [POD;Standard]] guard does not help ([E] is POD and
+   [D] is standard-layout), and neither would a per-class layout check on [E],
+   which has no members at all: the offending layout belongs to [D] while the
+   over-claiming rule is applied at [E].
+
+   The only reason this is not already exploitable is that there is no
+   [eval_o_base] (the base-class analogue of [eval_o_field]), so a base offset
+   cannot be evaluated at the interface level -- every concrete pointer model
+   does compute it ([simple_pointers_utils.o_base_off]). Adding [eval_o_base]
+   must therefore wait until this rule is fixed.
+
+   The fix is for a base subobject's extent to be the class's non-virtual
+   size (0 for an empty class) rather than [size_of], so the per-base entries
+   in [rss] cover that instead. Nothing in the workspace uses this rule today,
+   so it is commented out rather than weakened in place.
+
+   See https://github.com/SkyLabsAI/BRiCk/issues/310 for the follow-up work.
+
   Axiom struct_to_raw : forall cls st rss q,
     glob_def σ cls = Some (Gstruct st) ->
     st.(s_layout) ∈ [POD;Standard] ->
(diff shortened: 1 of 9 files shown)


rlepigre-skylabs-ai (approved):

rocq-skylabs-brick/theories/lang/cpp/logic/layout.v
```diff
@@ -26,7 +26,45 @@ Section with_Σ.
...
+   does compute it ([simple_pointers_utils.o_base_off]). Adding [eval_o_base]
+   must therefore wait until this rule is fixed.
+
+   The fix is for a base subobject's extent to be the class's non-virtual
+   size (0 for an empty class) rather than [size_of], so the per-base entries
+   in [rss] cover that instead. Nothing in the workspace uses this rule today,
+   so it is commented out rather than weakened in place.
+
+   See https://github.com/SkyLabsAI/BRiCk/issues/310 for the follow-up work.
+
```
Janno, line 50: That's too much slop text in the semantics for my liking. Could you minimize this to just the cpp code and the last 4 lemmas statements from https://github.com/SkyLabsAI/auto/pull/441? I think I would prefer that to this description here. It might even be enough to rewrite the paragraph in line 40 to basically just "the empy base class has the wrong size and we end up with both the base class and the E's first field claiming to own the first raw byte" or something like that.

Janno (dismissed):

rocq-skylabs-brick/theories/lang/cpp/logic/layout.v
pgiarrusso-sl: Cleaned up.

Janno: Much better! Thank you!

rocq-skylabs-brick/theories/lang/cpp/logic/layout.v
```diff
@@ -68,7 +94,7 @@ Section with_Σ.
       un.(u_trivially_destructible) ->
       cQp.frac q = 1%Qp ->
           type_ptrR (Tnamed cls)
-      |-- (Reduce (union_defR tblockR cls un q)) -* |={↑pred_ns}=> tblockR (Tnamed cls) q.
+      |-- (Reduce (union_defR tblockR cls un q)) -* |={↑pred_ns}=> tblockR (Tnamed cls) q. *)
```
simon-skylabs, line 98: To ensure that the axiom's statement remains type checked for future iterations, could we do one of the following?

1. turn the `Axiom` into a definition like `Definition disabled_axiom_<the_name> := <the statement>`
2. put the axiom in a module type that we don't instantiate.

I think 1. might be better.

jhaag-skylabs-ai (approved):

rocq-skylabs-brick/theories/lang/cpp/logic/layout.v
pgiarrusso-sl: I understand the idea but I don't think it's worthwhile. Going to merge this after rebase.

PR #313, opened 2026-09-09
expose `is_projection` to both Ltac and Ltac2

simon-skylabs:
The existing functions in Ltac and Ltac2, both named `is_proj`, only return true for primitive projections. This one takes non-primitive projections into account as well

### DIFF
--- a/ltac2-extra/theories/internal/ltac1.v
+++ b/ltac2-extra/theories/internal/ltac1.v
@@ -7,13 +7,14 @@
 
 Require Import skylabs.ltac2.extra.internal.init.
 Require Import skylabs.ltac2.extra.internal.constr.
+Require Import skylabs.ltac2.extra.internal.control.
 Require Import skylabs.ltac2.extra.internal.string.
 Require Import skylabs.ltac2.extra.internal.option.
 
 (** Minor extensions to [Ltac2.Ltac1] *)
 Module Ltac1.
   Import Ltac2 Init.
-  Export Ltac2.Ltac1.
+  Export Ltac2.Ltac1 Constr.
 
   (** [to_string v] attempts to build a string from an Ltac1 value, intended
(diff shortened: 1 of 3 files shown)


ltac2-extra/theories/internal/ltac1.v
```diff
@@ -32,4 +33,24 @@ Module Ltac1.
...
+                     | _ =>
+                         throw_invalid! "is_proj_ext: expecting a constant name but got %t" x
+                     end
+                 | None =>
+                     throw_invalid! "is_proj_ext: expecting constr but got %s" (Ltac1.tag_name ltac1_x)
+                 end in
+               if Constr.is_proj_ext x then () else Control.zero (Tactic_failure None)) in
+    tac x.
+
 End Ltac1.
```
Janno, line 56: I don't think this is the right module to put this. Ltac2.Ltac1 is explicitly for the FFI. This is here is just a client of the FFI.  I am not sure if we have an Ltac1 prelude somewhere; if so, that would be a better place.

Do you have an urgent use case for this in Ltac1? If not, I would just not define it at all. One more reason for people to switch to Ltac2..

Janno (changes requested):

ltac2-extra/theories/internal/ltac1.v
simon-skylabs: No that's true I don't. I think we could stand to make it easier to wrap Ltac2 into Ltac1 though. If we could use combinators, I'd like to be able to write something like `to_ltac1 (Arg.constr ==> Arg.ident ==> Ret.unit) my_ltac2_tactic` and trust that we'll get sensible error messages when arguments don't match up. What do you think and is this a good place to put this?

Janno: That would fit well in ltac1.v

PR #314, opened 2026-09-11
Draft: adapt specification language to allow the specification of postconditions for exception cases

simon-skylabs:
Step 1:
 * adjust definition of `WpSpec` to allow exception post condition terms

Step2:
 * define notation for new capabilities, in particular `\post_except` and `\strong_exception_safety`
   * No exceptions: when neither `\post_except` nor `\strong_exception_safety` are used in a spec, the exception-case post condition is taken to be `False`.
   * Strong exception safety: when `\strong_exception_safety` is used as the first keyword in a specification, `\pre{x} P x` is implicitly taken to mean `\with x \pre P x \post_except* P x`.
   * Weak exception safety: when `\post_except` is used without `\strong_exception_safety`, `\prepost{x} P x` is taken to mean `\pre{x} P x \post*  P x \post_except* P x` but for resources taken via `\pre`, the user is responsible for specifying how they are returned when an exception is raised.

## The syntax

There are two forms for `\post_except`:
  * `\post_except* P`: without consideration for the type of the exception object, `P` is expected to hold upon exceptional return.
  * `\post_except{p ty}[Vptr p, ty] P p ty`: `P` can perform a case distinction on type `ty` to specify a different postcondition for various types or assert that the type is among a certain list of allowed exception types. 

## An open problem

One open problem is how to specify the ownership of exception objects.

I am uncertain about whether `P` should specify the exception object.
 * Pro: it would be useful when we want to state that certain information exists in a specification object upon exceptional exit;
 * Con: if we want to state strong-ish exception safety, we may not want to list all available options because they may not be known.

One way through is consider that we can frame out the exception object. We may do that with an exception precondition:

```
\pre_except{p ty}[Vptr p, ty]  
  if bool_decide (ty = "std::logic_exception") then 
    Exists state, p |-> LogicException.R 1$m state 
  else emp
\post_except{p ty}[Vptr p, ty]  
  if bool_decide (ty = "std::logic_exception") then 
    Exists state, p |-> LogicException.R 1$m state ** ... 
  else emp
```

In effect, `pre_except` would be borrowing an exception object from the C++ abstract machine and the postcondition returns it.

### DIFF
--- a/rocq-skylabs-brick/theories/lang/cpp/specs/classy.v
+++ b/rocq-skylabs-brick/theories/lang/cpp/specs/classy.v
@@ -21,12 +21,47 @@ Require Import skylabs.lang.cpp.specs.spec_notations.
 Section with_prop.
   Context {PROP : bi} {spec_car : Type}.
 
+  (* IDEA: \except instead of \exceptPost
+     modes:
+     - noexcept
+
+     - weak safety
+       \pre{x y} P x y
+       \exceptPost Other x
+       \post{z} Q x y z
+
+     - strong safety
+       - \strong_safety
+         \pre{x} P x
+         \post{y} Q x y
+       - equivalent to
+         \pre{x} P x
+         \exceptPost P x
+         \post{y} Q x y
+     - \except{p}[Vptr p]
+   *)
+
+  (* Notation:
+     \post*         -- either normal exit or exception and normal exit
+     \exceptPost    -- only on exception
+     \post_all      -- exception and normal
+     \post_normal   -- normal exit
+
+     implicit \exceptPost False when?
+       - when no except postcondition is specified?
+   *)
+
   Class SpecGen : Type :=
     { add_pre : PROP -> spec_car -> spec_car
     ; add_post : PROP -> spec_car -> spec_car
     ; add_with : forall {T : Type@{universes.Quant}}, (T -> spec_car) -> PrimString.string -> dummy_prop -> spec_car
+
+    ; add_except_post : PROP -> spec_car -> spec_car
+      (* assumption: for exception safety, prepost adds an except post as well and normal post *)
+    ; add_post_all (P : PROP) (S : spec_car) : spec_car :=
+      add_post P (add_except_post P S)
     ; add_prepost (P : PROP) (S : spec_car) : spec_car :=
-      add_pre P (add_post P S)
+      add_pre P (add_post P (add_except_post P S))
     ; add_require (P : Prop) : spec_car -> spec_car :=
       add_pre (only_provable P)
     ; add_persist (P : PROP) : spec_car -> spec_car :=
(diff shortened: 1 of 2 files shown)


rocq-skylabs-brick/theories/lang/cpp/specs/wp_spec_compat.v
```diff
@@ -61,37 +61,45 @@ Section with_prop.
...
-    ; spec_internal_frame : forall args' P Q args K K',
-        (∀ r, K r -∗ K' r) ⊢ spec_internal args' P Q args K -∗ spec_internal args' P Q args K'
+      spec_internal : forall (acc_arg : list ARG) (acc_pre : list PROP)
+                        (acc_post : list (RESULT -> PROP))
+                        (acc_post_except : list (RESULT -> PROP)),
+        list ARG -> (RESULT -> PROP) -> (RESULT -> PROP) -> PROP
+    ; spec_internal_frame : forall args' P Q Q__except args K__normal K__normal' K__except K__except',
+        (∀ r, K__normal r -∗ K__normal' r) ∧
+        (∀ r, K__except r -∗ K__except' r)
+        ⊢ spec_internal args' P Q Q__except args K__normal K__except -∗ spec_internal args' P Q Q__except args K__normal' K__except'
```
pgiarrusso-sl, line 71: As mentioned, I'd rather change `RESULT` to a sum than complicate these statements...

rocq-skylabs-brick/theories/lang/cpp/specs/classy.v
```diff
@@ -21,12 +21,47 @@ Require Import skylabs.lang.cpp.specs.spec_notations.
...
+       - when no except postcondition is specified?
+   *)
+
   Class SpecGen : Type :=
     { add_pre : PROP -> spec_car -> spec_car
     ; add_post : PROP -> spec_car -> spec_car
     ; add_with : forall {T : Type@{universes.Quant}}, (T -> spec_car) -> PrimString.string -> dummy_prop -> spec_car
+
+    ; add_except_post : PROP -> spec_car -> spec_car
+      (* assumption: for exception safety, prepost adds an except post as well and normal post *)
```
pgiarrusso-sl, line 60:
```suggestion
      (* assumption: `\prepost` remains idiomatic for noexcept and strong exception safety, so it also adds an except post. *)
```

rocq-skylabs-brick/theories/lang/cpp/specs/classy.v
```diff
@@ -45,6 +80,12 @@ Section with_prop.
     ; post_ret : RESULT -> PROP -> post_car
     }.
 
+  Class WithExceptPost {RESULT : Type} `{!@WithPost RESULT} : Type :=
+    { except_car : Type
```
pgiarrusso-sl, line 84: Seems unused?

PR #317, opened 2026-09-15
binop/erase_qualifier changes and axioms + `forbid_raw` upstreaming

Janno:


jhaag-skylabs-ai (changes requested):
> [!NOTE]
>
> Opus 5 @ xhigh used for preliminary review against https://eel.is/c++draft/, cf. subsequent quote blocks; findings audited by hand afterwards.

**Bottom line up front:** (cc/ @pgiarrusso-sl )
1. `binop`/`erase_qualifier` changes seem sound today, but they may be unsound when rules are added for `Bdotp`/`Bdotip` or for pointer-to-member equality
2. there are known/subtle gaps between the notion of "similar types" and how `erase_qualifiers` is used in practice, though this doesn't seem to make BRiCk unsound today

I propose that we centrally document both of these (w/standard references and BRiCk permalinks, cf. below) before merging this PR. Overall I do like the changes in this PR and I'm in favor of merging once we document known limitations detailed below.

---

**Re. potential future soundness hazard with `binop`/`erase_qualifier` changes**

Using `erase_qualifier` in `eval_binop_{pure, impure}` seems justified by the standard:
> - [[basic.type.qualifier]/1](https://eel.is/c++draft/basic.types#basic.type.qualifier-1) — cv-variants "shall have the same representation and alignment requirements". So `sizeof`/`alignof` cannot distinguish them, which is all the pointer rules use. [[basic.type.qualifier]/3](https://eel.is/c++draft/basic.types#basic.type.qualifier-3) extends this through array element types.
> -  [[expr.add]/6](https://eel.is/c++draft/expr.add#6) — the UB condition for pointer arithmetic is phrased as: "if the expressions P or Q have type 'pointer to cv T', where T and the array element type are not [**similar**](https://eel.is/c++draft/conv.qual#def:similar_types), the behavior is undefined." Per [[conv.qual]/2](https://eel.is/c++draft/conv.qual#2), **similar** = same qualification-decomposition ignoring the `cv_i`. That is precisely erasure. The standard has already decided cv is immaterial here.
> - [[expr.add]/4](https://eel.is/c++draft/expr.add#4) & [[expr.add]/5](https://eel.is/c++draft/expr.add#4) — strides and `ptrdiff` are defined by array indices, i.e. `sizeof(T)`.
> - [[expr.eq]](https://eel.is/c++draft/expr.eq) — operands are brought to their composite pointer type by qualification conversions before comparison, so the result cannot depend on cv.

but our axioms are technically stronger than what the standard prescribes, though this doesn't seem to yield a soundness issue currently:
> `erase_qualifiers` descends further than [[conv.qual]/1](https://eel.is/c++draft/conv.qual#1)'s decomposition, which only goes through "pointer to", "pointer to member of class C of type", "array of N", "array of unknown bound of" — with the terminal `U` required to be the same type. BRiCk additionally recurses into:
> 1. `Tfunction` return and parameters ([syntax/types.v:444](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/syntax/types.v#L444)) — identifies `void(int*)` with `void(const int*)`. Those are unrelated types; `void(*)(int*)` and `void(*)(const int*)` are not similar and `==` between them is ill-formed.
> 2. `Tref`/`Trv_ref` — identifies `int&` with `const int&`.
> 3. `Tmember_pointer` — member-function cv-qualifiers ride on the implicit `this` parameter (`Tmember_func` prepends `Tptr ty`, [syntax/types.v:1263-1271](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/syntax/types.v#L1261-L1271)), so `void (C::*)() const` and `void (C::*)()` collapse.
>
> None is exploitable today: `size_of` is None for `Tfunction`, `Tmember_pointer` and `Tref`, so the ptr/int rules never fire on them, and the comparison axioms are already `∀ ty`-polymorphic. It becomes a live hazard the moment `eval_binop_impure` gains rules for `Bdotp`/`Bdotip` or for pointer-to-member equality, where [[expr.eq]/4](https://eel.is/c++draft/expr.eq#4) makes the identity of the member observable. Worth a comment on the axiom, or narrowing it to a "similar types" relation.

We can't easily narrow the axiom today since we haven't defined a "similar types" relation so we'll just need to document this.

---

**Re. `erase_qualifiers` vs. "similar types"**

> [Currently] there is no similar/similar_types definition anywhere in BRiCk. `erase_qualifiers t1 = erase_qualifiers t2` is the de facto stand-in, and it's already used that way in the places that are really about [[conv.qual]](https://eel.is/c++draft/conv.qual):
> - [semantics/cast.v:488](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/semantics/cast.v#L488) — `convert`'s pointer branch: `is_pointer from && bool_decide (erase_qualifiers from = erase_qualifiers to)` makes the conversion the identity on values. That is morally the qualification conversion of [[conv.qual]/3](https://eel.is/c++draft/conv.qual#3), which the standard licenses precisely between similar types.
> - [logic/expr.v:778](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/logic/expr.v#L778) — `classify_cast` decides `Nothing`/`AddConst`/`RemoveConst` by erasure-equality of the two `decompose_type` cores.
> - [logic/new_delete.v:254](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/logic/new_delete.v#L254) and [:996](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/logic/new_delete.v#L996) — comments quoting [[expr.delete]/3](https://eel.is/c++draft/expr.delete#3) and [[conv.qual]](https://eel.is/c++draft/conv.qual) verbatim, with an explicit admission that the approximation is unresolved: ["we might need to insert erase_qualifiers/normalize_type calls to reflect that the standard only calls for 'similarity' (which has a more nuanced consideration of cv-qualifiers)."](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/logic/new_delete.v#L1000-L1002)
>
> So the gap is known and documented, just never filled in. (`normalize_type`, [syntax/types.v:817-819](https://github.com/SkyLabsAI/BRiCk/blob/837f50d6614bb42f5589dd89e080a7aa0b625b3a/rocq-skylabs-brick/theories/lang/cpp/syntax/types.v#L817-L819), is a different normalization — it also does `to_arg_type` parameter decay — and is not similarity either.)
>
> Worth knowing [when we define "similar types" relation]: erasure-equality differs from [[conv.qual]/2](https://eel.is/c++draft/conv.qual#2) in both directions, not just the one [flagged above].
> - Too coarse (identifies types that are not similar): `erase_qualifiers` descends into `Tfunction` return/params, into `Tref`/`Trv_ref`, and into `Tmember_pointer`'s function type where member cv-qualifiers live on the synthesized `this` parameter. The standard's qualification-decomposition descends only through "pointer to", "pointer to member of class C of type", "array of N", "array of unknown bound of", and requires the terminal `U` to be the same type.
> - Too fine (separates types that are similar): [[conv.qual]/2](https://eel.is/c++draft/conv.qual#2) allows corresponding `Pᵢ` to be "array of `Nᵢ`" in one and "array of unknown bound of" in the other, so `int[5]` and `int[]` are similar — but `Tarray t 5` and `Tincomplete_array t` have different erasures. `Tvariable_array t e` is worse: the size expression `e` is left untouched by erase_qualifiers, so two VLA types differing only in cv inside `e` also stay distinct.

rocq-skylabs-brick/theories/lang/cpp/logic/heap_pred/prim.v
```diff
@@ -323,4 +323,50 @@ Section with_cpp.
...
+      bool_decide (ty ∈ [Tschar;Tushort;Tshort;Tint;Tuint;Tlong;Tulong;Tlonglong;Tulonglong;Tint128_t;Tuint128_t])
+    | Tptr _
+    | Tenum _
+    | Tchar_ _ (* this might change *)
+    | Tbool
+    | Tnullptr => true
+    | _ => false
+    end.
+
+  Definition forbids_raw (ty : type) : bool :=
```
pgiarrusso-sl, line 340: Looking at `forbids_raw_not_raw` and `has_int_type'`, is the `Tnum` clause basically just `ty <> Tuchar`?

Style nit: Lifting `bool_decide` out lets you use its theory before `destruct ty`.

```suggestion
  (** [forbids_raw_not_raw] shows soundness of [forbids_raw] (but not completeness). *)
  (* [allows_raw] is a more natural definition, but doing it completely requires
     that we give [has_type] relations for [Tmember_pointer] and such. *)
  Definition forbids_raw_unqualified (ty : type) : Prop :=
    match ty with
    | Tuchar => False
    | Tnum _ _ => True
    | Tptr _
    | Tenum _
    | Tchar_ _ (* this might change *)
    | Tbool
    | Tnullptr => True
    | _ => False
    end.

  Definition forbids_raw (ty : type) : Prop :=
```

pgiarrusso-sl: (This might also need a `Decision (forbids_raw_unqualified ty)` instance, but `rewrite /forbids_raw_unqualified; case_match; apply _` or `case: ty => /=; apply _` should do the job).

pgiarrusso-sl:
> * `binop`/`erase_qualifier` changes seem sound today, but they may be unsound when rules are added for `Bdotp`/`Bdotip` or for pointer-to-member equality

We could note a TODO but I don't know if `Bdotp`/`Bdotip` should go through `wp_operand_binop`.

> there are known/subtle gaps between the notion of "similar types" and how `erase_qualifiers` is used in practice, though this doesn't seem to make BRiCk unsound today

Most uses of `erase_qualifiers` are about converting to heap types:
```
(** ** Heap Types

    The C++ memory is typed, but loosely so with respect to cv-qualifiers.
    For example, the C++ runtime invariant does *not* guarantee that a value
    stored in a memory cell of type <<int*>> points to a memory cell that is
    mutable. Generally, this will be true, but constructs such as
    <<const_cast<>>> and casting through pointers explicitly circumvent this.
    To track this, the C++ abstract machine tracks locations with "heap types"
    which are [decltype]s with all qualifiers erased and top-level references
    normalized to [Tref].

    Concrete examples of the above are:
    - <<const int x = 1>>           -- [tptsto "int" 1$c _ 1]
    - <<const int* x = nullptr>>    -- [tptsto "int*" 1$m _ nullptr]
    - <<int* const x = nullptr>>    -- [tptsto "int*" 1$c _ nullptr]
    - <<volatile int x = 0>>        -- not represted using [tptsto]
    - <<volatile int* p = nullptr>> -- [tptsto "int*" 1$m _ nullptr]
    - <<int&& r = ..>>              -- [tptsto "int&" 1$m _ _]
 *)
```

rocq-skylabs-brick/theories/lang/cpp/logic/operator.v
```diff
@@ -264,7 +278,7 @@ Section with_Σ.
   #[local] Definition eval_ptr_int_op (bo : BinOp) (f : Z -> Z) : Prop :=
     forall w s p1 p2 o ty,
       is_Some (size_of σ ty) ->
-      p2 = p1 ,, _sub ty (f o) ->
+      p2 = p1 ,, _sub (erase_qualifiers ty) (f o) ->
```
pgiarrusso-sl, line 281: These changes seem fine — heap types for `_sub` seems appropriate.

I'm less sure on the new axioms.

pgiarrusso-sl: Worth getting these bugfixes in, since we have so many PRs in flight.

PR #319, opened 2026-09-16
fix: erase qualifiers in implicit array initialization

pgiarrusso-sl:


rocq-skylabs-brick/theories/lang/cpp/logic/expr.v
```diff
@@ -1832,7 +1812,7 @@ Module Type Expr.
         is_array_of aty ety ->
         get_default ety = Some default ->
         (base |-> type_ptrR (Tarray ety sz) -*
-         base |-> arrayR ety (primR ety 1$m) (replicateN sz default) -*
+         base |-> arrayR ety (primR (erase_qualifiers ety) 1$m) (replicateN sz default) -*
```
pgiarrusso-sl, line 1815: We _introduce_ `primR` ownership, and `primR`'s argument is supposed to always be qualifier-free.

I'm now wondering if `primR` should respect `erase_qualifiers` instead... @Janno ?

Janno: The comment on `primR` says:
```
  NOTE: Clients use [primR (erase_qualifiers ty)] but we do not bake
  [erase_qualifiers] in to simplify our automation.
```
I suspect this means that, with the current setup, the automation is not presented with a quotient because erasure happens explicitly and will be computed away before the automation continues.. at the cost of having to stick `erase_qualifiers` absolutely everywhere. 

I am honestly very torn on this. The automation does not make it easy to perform global rewriting so I understand the motivation. We don't even support global non-setoid rewriting right now. This might require some discussion to see if we can figure something out (perhaps more generally than just for `primR`).

pgiarrusso-sl: But then could we just keep `erase_qualifiers` in hints like today but add this axiom/derived rule?
```
    Axiom primR_erase: forall ty q p,
        primR ty q v -|- primR (erase_qualifiers ty) q v.
```

Then forgetting `erase_qualifiers` in axioms make no difference, and forgetting it in hints becomes just a completeness problem?

pgiarrusso-sl: Deferring to https://github.com/SkyLabsAI/BRiCk/issues/328.

Janno (approved):

PR #321, opened 2026-09-16
fix(logic/operator): remove unsound `eval_binop_impure_well_typed` in favor of type-constrained inversion on purity of eval_binop

jhaag-skylabs-ai:
Missing premises in introduction rules for `eval_binop_impure` allowed `eval_binop_impure_well_typed` to be used to derive `False`. This PR:
1. removes `eval_binop_impure_well_typed` + `eval_binop_impure_well_typed_pure`
2. adds `binop_needs_memory` so that `eval_binop` can be defined by discrimination on typing information instead of via logical disjunction
3. uses (2) to prove a stronger form of `eval_binop_well_typed`, sufficient & necessary for downstream clients


pgiarrusso-sl:
Thanks for this work! Since https://github.com/SkyLabsAI/BRiCk/pull/323 was approved and includes these commits, I'll close this one in favor of that one.

PR #326, opened 2026-09-21
Move the [cpp.prog] plugin to its own package.

rlepigre-skylabs-ai:
Less invasive version of https://github.com/SkyLabsAI/BRiCk/pull/249.


rocq-skylabs-brick/dune-project
```diff
@@ -37,12 +37,8 @@
...
   rocq-skylabs-iris
-  rocq-skylabs-cpp2v
-  ; this package mostly does not depend on cpp2v, except
-  ; that rocq-skylabs-brick/theories/lang/cpp/dune uses
-  ; cpp2v's hash in build parser/plugin/cpp2v.v.
-  ; TODO: make that a separate package.
   rocq-flocq
-  rocq-elpi))
+  rocq-elpi
+  (rocq-skylabs-brick-cpp :with-test)))
```
pgiarrusso-sl: Should this be
```suggestion
  (rocq-skylabs-brick-cpp-prog :with-test)))
```
and if yes, ~~why did CI pass?~~ Oh, CI passed because we don't test `opam install --with-test` (and we don't really care?)

We could even drop this until we change support policies, no strong opinions.
```suggestion
```

rlepigre-skylabs-ai: Good catch! Keeping the test dependency, but agreed we're not testing this in CI.

pgiarrusso-sl (approved):

